PR #50, opened 2019-07-18
fix: do not count on travis branch ci check

fntlnz:
As shown in picture, when an external contributor (or anyone using a fork) does a PR the Travis branch check does not trigger so the PR can never be merged.

I disabled this check on Prow with this PR.

Also, that check just gets reported when the other Travis check passes (when it works) so it's not useful to have it in place.

Screenshot 2019-07-18 at 13 51 09

Signed-off-by: Lorenzo Fontana <lo@linux.com>

### DIFF
--- a/prow/config/configs.yaml
+++ b/prow/config/configs.yaml
@@ -20,7 +20,6 @@ branch-protection:
         falco:
           required_status_checks:
             contexts:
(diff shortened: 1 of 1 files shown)


fntlnz:
@leodido PTAL

PR #103, opened 2020-03-16
new: add Flatcar stable support

mazzy89:
Signed-off-by: Salvatore Mazzarino <dev@mazzarino.cz>

Flatcar brought the CoreOS legacy and it will be the replacement.

Flatcar releases the OS into different channels flatcar-linux.org/releases

I'm targetting here Flatcar stable.

### DIFF
--- a/driverkit/config/dev/flatcar_stable_4.19.106-flatcar.yaml
+++ b/driverkit/config/dev/flatcar_stable_4.19.106-flatcar.yaml
@@ -1,0 +1,5 @@
+kernelrelease: 4.19.106-flatcar
+target: flatcar-stable
+output:
+    module: output/dev/falco_flatcar_stable_4.19.106-flatcar.ko
+    probe: output/dev/falco_flatcar_stable_4.19.106-flatcar.o


driverkit/config/dev/flatcar_stable_4.19.106-flatcar.yaml
```diff
@@ -0,0 +1,5 @@
+kernelrelease: 4.19.106-flatcar
+target: flatcar-stable
```
alban: Do the targets in driverkit need to be different for each flavor of Flatcar? There is currently 4 channels: stable, beta, alpha, edge.
I see that for CentOS, there is a single CentOS target, and the implementation is able to try different URLs for different releases. Is it possible to do something like that with an unique target "flatcar"?

mazzy89: Yeah, it makes sense. if you are already spotted a pattern that the guys put in place then let's follow for flatcar too.

mazzy89:
@leogr some nice info here:

```sh
➜ ./driverkit docker -c ./vanilla.yaml --timeout=300
INFO[0000] using config file                             file=./vanilla.yaml
INFO[0000] driver building, it will take a few seconds to complete  processor=docker
INFO[0089] kernel module available                       path=/tmp/falco-vanilla.ko
INFO[0089] eBPF probe available                          path=/tmp/falco-vanilla.o
```

```
➜ ls -lh /tmp/falco*
-rw-r--r--  1 salvatoremazzarino  wheel   3.2M Mar 17 16:35 /tmp/falco-vanilla.ko
-rw-r--r--  1 salvatoremazzarino  wheel   3.1M Mar 17 16:35 /tmp/falco-vanilla.o
```

with the following config:

```yaml
kernelrelease: 4.19.96-flatcar
kernelversion: 1
target: vanilla
output:
  module: /tmp/falco-vanilla.ko
  probe: /tmp/falco-vanilla.o
driverversion: 0de226085cc4603c45ebb6883ca4cacae0bd25b2
kernelconfigdata: ...
```

Any further inputs here?

driverkit/config/dev/flatcar_stable_4.19.106-flatcar.yaml
leogr: Hi @mazzy89 
I believe we need [this](https://github.com/falcosecurity/driverkit/issues/21#issuecomment-601117055) done before. 
Alternatively, if the configuration works just with `target: vanilla`, we could use it now, and later we will add the flatcar target as further optimization.
Let me know if you need any help with that!

mazzy89: Hi @leogr thank you for the input here.
I forgot to ask last time. how can we make sure that the `target: vanilla` used to build the driver for flatcar actually work?

leogr: Firstly, we should make sure it works by [manual installation](https://github.com/falcosecurity/driverkit/issues/23). Have you tried?

If so - and if I am not wrong - it should also work on all kernels with the same `kernelconfigdata`.

mazzy89: This test for me then would require a bit more of time because being in a cluster i would need to change the userdataa and then make an entire rollout

fntlnz:
Closing this since to add flatcar we need to really take care about Driverkit and add the distribution there first as per https://github.com/falcosecurity/driverkit/issues/21

Thanks for looking into this @mazzy89 - if you want to do further development and need assistance please let me know, I'd be happy to help.

/close

PR #131, opened 2020-05-20
contrib -> evolution

leodido:
This PR is related to https://github.com/falcosecurity/contrib/pull/24 about the renaming of the `contrib` repo to `evolution`.

Furthermore, it adds custom `kind/sandbox` etc. labels for it (as per https://github.com/falcosecurity/contrib/issues/14),

while at the same moment it also deletes custom `kind/pr` and `kind/issues` labels for https://github.com/falcosecurity/community (which ISSUE_TEMPLATEs do not use anymore).

### DIFF
--- a/prow/config/plugins.yaml
+++ b/prow/config/plugins.yaml
@@ -10,7 +10,7 @@ approve:
       - falcosecurity/cloud-native-security-hub-frontend
       - falcosecurity/cloud-native-security-hub-backend
       - falcosecurity/community
-      - falcosecurity/contrib
+      - falcosecurity/evolution
       - falcosecurity/driverkit
       - falcosecurity/event-generator
       - falcosecurity/falco
@@ -43,6 +43,10 @@ label:
   additional_labels:
     - kind/pr
     - kind/issue
+    # Custom kind/* lables for evolution repository
+    - kind/sandbox
+    - kind/incubation
+    - kind/officialsupport
 
 lgtm:
   - repos:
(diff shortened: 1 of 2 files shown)


prow/config/plugins.yaml
```diff
@@ -40,8 +40,10 @@ goose:
 
 label:
   additional_labels:
-    - kind/pr
-    - kind/issue
+    # Custom kind/* labels for evolution repository
+    - kind/sandbox
```
leodido, line 44: @kris-nova @leogr this PR is blocked from two things:

- [ ] `contrib` rename to `evolution`
- [ ] decide what label do we want to identify/automate the level (and create ISSUE_TEMPLATEs accordingly)

I can proceed with the rename, but then I would not be able to merge this PR (which instructs prow about the name change) because blocked by point 2.

leogr: As discussed [here](https://github.com/falcosecurity/contrib/issues/19#issuecomment-628548388), I thought we already decide about the following labels:
- `kind/sandbox`
- `kind/incubating`
- `kind/officialsupport`
(they just match the status name, as documented [here](https://github.com/falcosecurity/contrib/pull/24/files#diff-04c6e90faac2675aa89e2176d2eec7d8))

Regarding the ISSUE_TEMPLATEs, I can fix them.

leogr: - [x] [labels added](https://github.com/falcosecurity/contrib/labels)
- [x] Issue templates added to [contrib#24](https://github.com/falcosecurity/contrib/pull/24)

PR #150, opened 2020-07-09
override (and disable) the pull request reviews for falcosecurity/charts#gh-pages branch

leodido:
In an attempt to let @poiana push to `gh-pages` branch of `falcosecurity/charts` as per @leogr needs about the charts release process.

The idea here is to remove the need of (at least 1) reviewer on pull-requests.

Note that the `gh-pages` branch is still protected but users in `maintainers` and bot dedicated GitHub teams can still push to it.

As per branchprotector [docs](https://github.com/kubernetes/test-infra/tree/8c8546d74168a57d0a3d22b17d98eb09bbf70601/prow/cmd/branchprotector#scope) bool/int values should override (not merge) with parent hierarchical YAML keys.

Signed-off-by: Leonardo Di Donato <leodidonato@gmail.com>

### DIFF
--- a/prow/config/configs.yaml
+++ b/prow/config/configs.yaml
@@ -1,6 +1,6 @@
 branch-protection:
   enforce_admins: true # rules apply to admins too!
-  restrictions:
+  restrictions: # restrict who can push
     teams: ["maintainers", "machine_users"]
   required_pull_request_reviews:
     dismiss_stale_reviews: true # automat dismiss old reviews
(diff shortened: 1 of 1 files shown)


prow/config/configs.yaml
```diff
@@ -33,6 +33,9 @@ branch-protection:
               protect: true
             gh-pages:
               protect: true
+              required_pull_request_reviews:
```
fntlnz, line 36: Can you add a comment here to explain to the future us why this has been done?
I don't want to find myself removing it in two months

leogr, line 36:
```suggestion
              required_pull_request_reviews: # disable PR reviews since our bot needs to push while the CI workflow is running
```

PR #169, opened 2020-09-02
Automatic drivers cleanup

leodido:
This PR introduces:

- a script into `driverkit/utils` that (given the bintray secret plus a list of driver versions to keep) queries bintray API to obtain all the existing driver versions and filter them out by keeping only the two driver versions specified as input
- a make command that calls the above script specifying the bintray secret and passes to it the driver versions currently in the `driverkit/config` directory
- adds to the CI a job (`drivers/cleanup`) that puts all the pieces together and runs on every merge into master

### DIFF
--- a/driverkit/utils/cleanup
+++ b/driverkit/utils/cleanup
@@ -14,15 +14,15 @@ filterout_versions() {
         printf "%s\t%s\n" "$i" "${all[$i]}"
         if [[ ! " ${keep[*]}" =~  ${all[$i]} ]]; then
             printf " \tremoving..."
-            yes N | JFROG_CLI_LOG_LEVEL=DEBUG jfrog bt vd --user "${user}" --key "${pass}" "falcosecurity/driver/kernel-module/${all[$i]}" # --quiet (remove yes)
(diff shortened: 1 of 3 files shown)


driverkit/utils/cleanup
```diff
@@ -0,0 +1,55 @@
...
+    IFS=$'\n' read -r -d '' -a all < <(curl -s -u "${user}:${pass}" --header "Content-Type: application/json" https://api.bintray.com/packages/falcosecurity/driver/kernel-module | jq -r '.versions | .[]')
+}
+
+filterout_versions() {
+    for i in "${!all[@]}";
+    do
+        printf "%s\t%s\n" "$i" "${all[$i]}"
+        if [[ ! " ${keep[*]}" =~  ${all[$i]} ]]; then
+            printf " \tremoving..."
+            JFROG_CLI_LOG_LEVEL=DEBUG jfrog bt vd --user "${user}" --key "${pass}" "falcosecurity/driver/kernel-module/${all[$i]}" --quiet
```
leodido, line 17: Here we could download (or directly move the driver) to S3 for archivation as per proposal in https://github.com/falcosecurity/falco/pull/1375.


Waiting for feedback.

Should we block this task until we have S3?
Should we implement the archivation step in a second phase?

leodido:
Until I do some other checks

/hold

fntlnz (approved):

leogr (approved):

PR #190, opened 2020-10-05
update(driverkit/config): Driver support for newer gcp kernel

singhswg:


leodido (commented):
Thanks for submitting this!

Anyways, have you tried to build this configuration locally by using the `falcosecurity/driverkit` tool?

Because as far as I remember driverkit doesn't support it yet.

Furthermore would you please sign-off your commit? :)

singhswg:
@leodido The **contributing guide** doesnt open for me but i did try to push an empty change with `SIgned off by` but looks like it didnt work. Also i tried to test the configuration locally using `driverkit` tool but it failed - 

```
go run main.go docker --output-module /tmp/falco-1.ko --kernelversion=25 --kernelrelease=5.4.0-1025-gcp --driverversion=dev --target=ubuntu-generic
INFO driver building, it will take a few seconds   processor=docker
FATA exiting                                       error="specific kernel headers not found"
exit status 1
```

This worked though -

```
go run main.go docker --output-module /tmp/falco.ko --kernelversion=81 --kernelrelease=4.15.0-72-generic --driverversion=dev --target=ubuntu-generic
INFO driver building, it will take a few seconds   processor=docker
INFO kernel module available                       path=/tmp/falco.ko 
```

So you are probably right when you say `driverkit doesn't support it yet.`

fntlnz:
@singhswg as @leodido was saying the command for gcp can't work because Driverkit, the tools that does the builds does not have support for it. Every supported distribution has a specific file here in this folder https://github.com/falcosecurity/driverkit/tree/master/pkg/driverbuilder

Are you on GKE? Usually GKE users use the eBPF probe instead of the module and that one builds on the fly and does not need a prebuilt module usually. Are you running into issues with that?

maxgio92:
Hi @singhswg, did you get a chance to try the eBPF probe instead of the prebuilt module?
Thank you

PR #204, opened 2020-11-11
Clean branch prow changes for migration

jonahjon:


fntlnz (approved):

.circleci/config.yml
```diff
@@ -240,6 +240,44 @@ jobs:
...
               make cleanup_s3
+  "test-infra/scan/terraform":
+    docker:
+      - image: accurics/terrascan:latest
+    steps:
+      - checkout
+      - run:
+          name: Scan terraform code
+          command: |
+            #terrascan scan -d ./config/clusters -i terraform -t aws
```
leodido, line 251: Why is this commented?

leodido, line 251:
```suggestion
            terrascan scan -d ./config/clusters -i terraform -t aws
```

Makefile
```diff
@@ -0,0 +1,56 @@
+SHELL := /bin/bash
+
+
+.PHONY: oauth-token hmac-token github-oauth-config cookie plugins update-config
+
+### Prow Compenents
```
leodido, line 6:
```suggestion
### Prow Components
```

Typo

Makefile
```diff
@@ -0,0 +1,56 @@
...
+
+1password-local:
+	export $$(xargs <.env) && ./tools/deploy_prow.sh
```
leodido, line 9: Just a question: 
why this make command also deploys prow?

tools/deploy_prow.sh
```diff
@@ -0,0 +1,56 @@
...
+  echo "Launching Prow microservices" 
+  launchProw
+  echo "All done!"
+}
+
+function updateKubeConfig() {
+  aws eks --region ${ZONE} update-kubeconfig --name ${CLUSTER}-test-infra
+}
+
+function launchConfig(){
```
leodido, line 25: Why this script does not use the already existing makefile commands? 

I would stick with one single strategy for deploying it: bash or make commands

leodido (commented):
Just some questions, otherwise LGTM

.circleci/config.yml
jonahjon: This is a max question he wrote this job, @maxgio92

Makefile
jonahjon: This was for local testing, so it would pull the configmaps and deploy all the peices without failing... it probably can removed for now if wanted.

tools/deploy_prow.sh
jonahjon: The idea was that the scripts were for CircleCI, and the make commands were quick hacks locally for testing. Future we could move all to make, but it would require to testing.

.circleci/config.yml
maxgio92: It remained as a placeholder as it didn't work. We can remove it and add it in a separate PR. WDYT?

tools/deploy_prow.sh
leodido: Can't we reuse the make commands in that bash script?

jonahjon: We can, except in the case we need to remake the entire cluster the commands will fail since are replacement commands.  `kubectl create secret generic hmac-token --from-literal=hmac="$$(./tools/1password.sh -d config/prow/hmac-token)" --dry-run -o yaml | kubectl replace secret hmac-token -f -`

jonahjon: Moving it all to make based for CircleCI we can try in another PR, would require some time for testing, WDYT?

leodido: 👌

.circleci/config.yml
maxgio92: I fixed it, let's see if it works now. Otherwise we can remove it for now I think and open a separate PR for it

maxgio92: Ok, it works. It now prevents to deploy "unsecure" infrastructure code

leodido (approved):

fntlnz (approved):

PR #209, opened 2020-11-17
lifecycle bot

### LINKED ISSUE #40: Lifecycle bot
Create a stale "bot" using a set of prowjobs in order to automatically tag the lifecycle phase of issues.

Currently, we have the stale bot (a GitHub action) that checks issues and pings authors if they haven't been updated in a while.
The way it's configured and the way it works is not clear neither transparent to our users.

In fact, we experienced some complaints in the past about it.

We should remove such bot and implement a better one to automate the lifecycle of issues (and maybe pull requests too) of the whole falcosecurity organization.

## Lifecycle

- frozen: issue or PR should not be autoclosed due to rottenness
- stale: issue or PR open with no activity (90days) and has become stale
- rotten: 30days have passed since the issue or the PR was marked as stale, it will be automatically closed soon
- close: after other 30days the rotten issue is automatically closed

Thus, 90+30+30 = an issue would be closed without _any_ activity after 150 days.

Using `/lifecycle frozen` an issue will be excluded from this process.

/cc @falcosecurity/test-infra-maintainers

leodido:
Fixes #40 (where you can find detailed information about the resulting behavior of these 2 periodic `ProwJob`s.)

The lifecycle transitions are:

1. issue or PR becomes stale after 90 days of inactivity
2. stale issue or PR becomes rotten after 30 days of inactivity
3. rotten issues close after 30 days of inactivity

WIP because some points are still open:

- [ ] should we build and publish our `commenter` image?
- [ ] disable existing GitHub stale bot action

### DIFF
--- a/config/jobs/lifecycle-bot/periodic-close.yaml
+++ b/config/jobs/lifecycle-bot/periodic-close.yaml
@@ -1,0 +1,39 @@
+periodics:
+  - name: periodic-close
+    interval: 1h
+    decorate: true
+    annotations:
+      description: Closes rotten issues after 30d of inactivity
+    agent: kubernetes
+    spec:
+      containers:
+        - image: gcr.io/k8s-prow/commenter:v20201116-dbd60ab192
+          command:
+            - /app/robots/commenter/app.binary
+          args:
+            - |-
+              --query=org:falcosecurity
+              -label:lifecycle/frozen
+              label:lifecycle/rotten
+            - --updated=720h
+            - --token=/etc/github/oauth
+            - |-
+              --comment=Rotten issues close after 30d of inactivity.
+
+              Reopen the issue with `/reopen`.
+
+              Mark the issue as fresh with `/remove-lifecycle rotten`.
+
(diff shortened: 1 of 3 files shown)


jonahjon:
lgtm

jonahjon:
> should we build and publish our commenter image?

Is there any benefit to this?

> disable existing GitHub stale bot action

Not sure how this is currently setup, but seems like a good idea.


The prow job's look good either way, maybe moving interval up slightly from 1h to like 6h would be the only change I'd make.

jonahjon (approved):

leodido:
> Is there any benefit to this?

IMHO, not in this case :)
But I wanted your opinion!

> Not sure how this is currently setup, but seems like a good idea.

It's a Github App (to be precise) in the settings of the whole organisation (if I don't remember it wrong).
We need to think how to let all the test-infra-maintainers handle such things.

> The prow job's look good either way, maybe moving interval up slightly from 1h to like 6h would be the only change I'd make.

Thanks Jonah!

fntlnz (approved):

PR #210, opened 2020-11-17
Add build-drivers jobs for all distros

maxgio92:
Add drivers build Prow jobs for all Linux distributions

### DIFF
--- a/config/jobs/build-drivers/build-drivers.yaml
+++ b/config/jobs/build-drivers/build-drivers.yaml
@@ -6,6 +6,7 @@ presubmits:
     agent: kubernetes
     branches:
       - ^master$
+    run_if_changed: '^driverkit\/(?!README.md)'
     spec:
       containers:
       - command:
@@ -24,6 +25,7 @@ presubmits:
     agent: kubernetes
     branches:
       - ^master$
+    run_if_changed: '^driverkit\/(?!README.md)'
     spec:
       containers:
       - command:
@@ -42,6 +44,7 @@ presubmits:
     agent: kubernetes
     branches:
       - ^master$
+    run_if_changed: '^driverkit\/(?!README.md)'
     spec:
       containers:
       - command:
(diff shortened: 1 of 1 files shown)


config/jobs/build-drivers/build-drivers.yaml
```diff
@@ -36,6 +36,78 @@ presubmits:
         imagePullPolicy: Always
         securityContext:
           privileged: true
+  - name: build-drivers-centos-presubmit
+    decorate: true
+    skip_report: false
+    agent: kubernetes
```
jonahjon, line 42: let's add a run if changed section here on all of these presubmits, so it doesnt run a bunch of 3Hour jobs on every PR. Here's a few examples of how to format that. 

for just driverkit config
`run_if_changed: '^driverkit/config/'`

or for config and utils
`run_if_changed: ^(driverkit/config/|driverkit/utils/)`

jonahjon (changes requested):

config/jobs/build-drivers/build-drivers.yaml
maxgio92: great idea! let's do it

maxgio92: maybe `driverkit/config/` is enough, WDYT @leodido ?

jonahjon: I'm cool with whatever.

jonahjon: My regex isn't the best.... lol will this will run on anything in `driverkit/` except the README?

jonahjon (approved):

jonahjon:
One potential issue I see is that this is a very resource intensive job. If all jobs start at the exact same moment it may cause them to be super slow/not finish. Let's wait and see lol

config/jobs/build-drivers/build-drivers.yaml
maxgio92: yess :)

maxgio92: Maybe regex negative look-ahead is not supported by Golang, I think we can let `checkconfig` find out :)

PR #216, opened 2020-11-18
Fix drivers build jobs triggers

maxgio92:
We need to set a proper changes trigger for the drivers build jobs.

(Something to consider for the regex format in Prow job's `run_if_changed` is that negative look aheads are not supported in Golang).

### DIFF
--- a/config/jobs/build-drivers/build-drivers.yaml
+++ b/config/jobs/build-drivers/build-drivers.yaml
@@ -6,7 +6,7 @@ presubmits:
     agent: kubernetes
     branches:
       - ^master$
-    run_if_changed: '^driverkit\/(?!README.md)'
+    run_if_changed: '^driverkit\/'
     spec:
       containers:
       - command:
@@ -25,7 +25,7 @@ presubmits:
     agent: kubernetes
     branches:
       - ^master$
-    run_if_changed: '^driverkit\/(?!README.md)'
+    run_if_changed: '^driverkit\/'
     spec:
       containers:
       - command:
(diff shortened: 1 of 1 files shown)


maxgio92:
IMHO it can be a starting point also considering the `/driverkit/README.md` updates in the trigger. WDYT?

leodido (approved):

maxgio92:
I unset the status setting on GitHub ([`skip_report`](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#posting-github-status-contexts)), as the builds require a lot of time and they are preventing [tide ](https://github.com/kubernetes/test-infra/tree/master/prow/tide)to merge PRs.
We need first to measure and monitor resource usage and setup resources requests/limits on each job.

jonahjon:
> I unset the status setting on GitHub ([`skip_report`](https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md#posting-github-status-contexts)), as the builds require a lot of time and they are preventing [tide ](https://github.com/kubernetes/test-infra/tree/master/prow/tide)to merge PRs.
> We need first to measure and monitor resource usage and setup resources requests/limits on each job IMHO. WDYT?

I think this is a good approach.... I worry about having 6 heavy jobs with the same interval `48h` all DDOS'ing the cluster every 2 days too. Staggering jobs, or add limits + autoscaling seems like good choices.

jonahjon:
+1 on the `skip_report` too

maxgio92:
I see you left review comments @leodido but it appears they aren't

leodido (approved):
As soon we get the skinny on how the `run_if_changed` expressions really work, we'll exclude the changes to `driverkit/README.md` file to trigger these jobs.
But not a top priority rn.

Thus LGTM!

PR #225, opened 2020-11-25
Update driver configurations (2020-11-25)

leodido:
Prebuilt drivers update

Also configures 5c0b863ddade7a45568c0ac97d037422c9efb750 driver version (for the upcoming Falco 0.27 release)

PR #237, opened 2020-12-02
Adding trigger, removing old prow config and circleCI driverkit jobs

jonahjon:

- Added trigger to test-infra, and added comments for plugin explanation
- Removed circleCI jobs for driverkit to help PR #225 
- Removed old prow config/plugins files to reduce confusion


config/plugins.yaml
```diff
@@ -244,6 +244,26 @@ triggers:
   only_org_members: true
 
 plugins:
+  falcosecurity/test-infra:
```
leodido, line 247: Can we maintain these keys in alphabetical order, please?

WDYT?

config/config.yaml
```diff
@@ -188,14 +188,6 @@ branch-protection:
           branches:
             master:
               protect: true
-          required_status_checks:
```
leodido, line 191: Since we're here... :)

Should we insert some ProwJob as a `required_check` to avoid PRs getting merged without checks other than approval?

leodido (commented):
Left two comments/questions

Thanks 🙏

config/plugins.yaml
jonahjon: Sure, I put it at the top since it had all the comments so people could find it easier. Would you prefer switching back?

config/config.yaml
jonahjon: I think at least `check-prow-config` would be a good idea. I can go ammend that

jonahjon: Added 3 here

```
          required_status_checks:
            contexts:
              - "ci/circleci: deploy/terraform"
              - "ci/circleci: deploy/prow"
              - "check-prow-config"
```

WDYT?

config/plugins.yaml
jonahjon: Fixed

config/config.yaml
leodido: 💯

leodido (approved):

PR #248, opened 2020-12-07
Prowjob local testing

maxgio92:
This is a work in progress on refactoring and fixing the local testing environment.


config/jobs/driverkit/driverkit-test.yaml
```diff
@@ -0,0 +1,21 @@
+presubmits:
```
jonahjon, line 1: We should double check if we want this job to be a local job or not. All `.yaml` files will get merged into real prowjobs.

maxgio92: Completely agree.
For conciseness I removed this driverkit-itself build jobs (not `build-drivers`), as it's out of scope of this PR IMHO.

jonahjon (approved):

PR #254, opened 2020-12-14
Filter driver build targets also per kernel version

### LINKED ISSUE #250: Split driverkit targets per Kernel per Distribution
**Motivation**

The [driverkit build jobs](https://github.com/falcosecurity/test-infra/tree/master/config/jobs/build-drivers) could be more atomic.

**Feature**

Enable specific `Make` targets for each Linux kernel version for each Linux distribution.
E.g.:
```shell
make -e TARGET_DISTRO=ubuntu-generic -e TARGET_KERNEL=5.9
```

**Alternatives**

Split `Make` Linux distribution specific target into slices with equal driver count.
E.g.:
```shell
make -e TARGET_DISTRO=ubuntu-generic -e TARGET_SLICE=1 # first <driver-count> slice
```

**Additional context**

Driverkit builds for Ubuntu Generic take now more than 12 hours.

maxgio92:
## Related issues
Closes #250.

## New features

As for the related issue this PR enables to further filter Falco driver versions to pre-build by specifying both:
- Linux distribution
- Linux version

by enabling the `specific_target` Make target to accept a new parameter: `TARGET_KERNEL`.

This parameter as for `TARGET_DISTRO` is read from the environment and it accetps a Linux version in `<kernel_version>.<major_version>.<minor_version>` format.

For instance:

```bash
make -e TARGET_DISTRO=centos -e TARGET_KERNEL=3.10.0 specific_target
[...]
```

It introduces also a Make target `count_specific_target` to count the result of the filter as for the `TARGET_*` variables set.
For instance:
```bash
make -e TARGET_DISTRO=centos -e TARGET_KERNEL=3.10.0 count_specific_target
261
```

### Breaking changes
None.

### DIFF
--- a/driverkit/Makefile
+++ b/driverkit/Makefile
@@ -5,16 +5,14 @@ DRIVERKIT ?= $(shell which driverkit)
 CONFIGS := $(wildcard config/*/*.yaml)
 VERSIONS := $(patsubst config/%,%,$(sort $(dir $(wildcard config/*/))))
 VERSIONS := $(VERSIONS:/=)
+TARGET_VERSION ?= *
 TARGET_DISTRO ?= *
 TARGET_KERNEL ?= *
 S3_DRIVERS_BUCKET ?= "falco-distribution/driver"
 
 all: $(patsubst config_%,%,$(subst /,_,$(CONFIGS)))
 
-specific_target: $(patsubst config_%,%,$(subst /,_,$(wildcard config/*/${TARGET_DISTRO}_${TARGET_KERNEL}-*)))
-
-count_specific_target:
-	@echo $(words $(patsubst config_%,%,$(subst /,_,$(wildcard config/*/${TARGET_DISTRO}_${TARGET_KERNEL}-*))))
+specific_target: $(patsubst config_%,%,$(subst /,_,$(wildcard config/${TARGET_VERSION}*/${TARGET_DISTRO}_${TARGET_KERNEL}-*)))
 
 prepare: $(addprefix prepare_,$(VERSIONS))
 publish: $(addprefix publish_,$(VERSIONS))
@@ -85,7 +83,7 @@ $(foreach VERSION,$(VERSIONS),\
 
 .PHONY: clean
 clean:
-	@find output -type f -not -name '.gitignore' -delete
+	find output/ -not -name '.gitignore' -not -name 'output' -delete
 
 stats:
 	@utils/driverstats $(CONFIGS)
--- a/driverkit/Makefile
+++ b/driverkit/Makefile
@@ -86,4 +86,4 @@ clean:
 	find output/ -not -name '.gitignore' -not -name 'output' -delete
 
 stats:
-	@utils/driverstats $(CONFIGS)
+	@utils/driverstats
--- a/driverkit/Makefile
+++ b/driverkit/Makefile
@@ -5,12 +5,16 @@ DRIVERKIT ?= $(shell which driverkit)
 CONFIGS := $(wildcard config/*/*.yaml)
 VERSIONS := $(patsubst config/%,%,$(sort $(dir $(wildcard config/*/))))
 VERSIONS := $(VERSIONS:/=)
-TARGET_DISTRO ?=
+TARGET_DISTRO ?= *
+TARGET_KERNEL ?= *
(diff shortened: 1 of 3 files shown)


driverkit/Makefile
```diff
@@ -5,12 +5,16 @@ DRIVERKIT ?= $(shell which driverkit)
 CONFIGS := $(wildcard config/*/*.yaml)
 VERSIONS := $(patsubst config/%,%,$(sort $(dir $(wildcard config/*/))))
 VERSIONS := $(VERSIONS:/=)
-TARGET_DISTRO ?=
+TARGET_DISTRO ?= *
```
leodido, line 8: Not sure we need `TARGET_DISTRO` to equal to `*` by default.

In case you need to build all you can use `make all`

driverkit/Makefile
```diff
@@ -5,12 +5,16 @@ DRIVERKIT ?= $(shell which driverkit)
...
+TARGET_KERNEL ?= *
 S3_DRIVERS_BUCKET ?= "falco-distribution/driver"
 
 all: $(patsubst config_%,%,$(subst /,_,$(CONFIGS)))
 
-specific_target: $(patsubst config_%,%,$(subst /,_,$(wildcard config/*/${TARGET_DISTRO}_*)))
+specific_target: $(patsubst config_%,%,$(subst /,_,$(wildcard config/*/${TARGET_DISTRO}_${TARGET_KERNEL}-*)))
```
leodido, line 14: Just an idea here, Max.

What if we make this filter broader? I mean, something like `TARGET_KERNEL=5.x` would be very cool and needed as filter.

Consider we already have `make <driver_versiion>_<distro>_<kernel>.yaml` make commands generated.

Run this to list all the make commands available:

```console
make -qp | awk -F':' '/^[a-zA-Z0-9][^$#\/\t=]*:([^=]|$)/ {split($1,A,/ /);for(i in A)print A[i]}' | sort -u
```

driverkit/Makefile
```diff
@@ -5,12 +5,16 @@ DRIVERKIT ?= $(shell which driverkit)
...
+
+count_specific_target:
```
leodido, line 16: This was already done by me Max :)

Below, there's the `make stats` command (which can be further parametrized maybe...).

Or you can directly use the script it calls.

Eg.

```console
$ ./utils/driverstats config/2aa88dcf6243982697811df4c1b484bcbe9488a2/*.yaml
# eBPF probes     191
# kernel modules  80
```

leodido (changes requested):
Left some comments Max!

driverkit/Makefile
maxgio92: I was thinking about it in case we want to filter only by kernel version, but maybe it's not useful?

driverkit/Makefile
maxgio92: I like the idea! I'll try to do it

driverkit/Makefile
maxgio92: Thanks. I think we can remove it, honestly I used it to test the changes on `specific_target` filter, and I left it just in case :D

driverkit/Makefile
leodido: It works Max!

Also added the same filtering feature for Falco driver versions.

driverkit/Makefile
leodido: Working on improviong/parametrizing `make stats`.

After that I'll consider this done and we can merge it!

leodido: Pushed something ;)

leodido (approved):

jonahjon:
Is there anyway to test this with a job?

driverkit/Makefile
maxgio92: Thank you

driverkit/Makefile
maxgio92: Love the hashmap simulation <3 Great work!

leodido:
@jonahjon what you exactly would like to test?

I see some options:
- ProwJobs that build drivers by kernel release majors (eg., "3.*", "4.*", "5.*")
- ProeJob that calls `make stats` with various combinations (varying TARGET_DISTRO and TARGET_KERNEL) and creates an artifact we could publish on the website (maybe on a `driverstats` page)

PR #256, opened 2020-12-15
Update drivers (2020-12-15)

fntlnz:
Updating the drivers...


leodido (approved):

fntlnz:
I think that this is failing because the copy command tries to copy bpf probes even on configurations that don't build them. I'll try to come up with a fix.

leodido:
Can we close this in favor of #260 ?

PR #267, opened 2020-12-22
Porting prow/update-jobs to Go modules

leodido:


jonahjon (approved):

jonahjon:
Although the make-jobs image and entrypoint might need to be updated. Can address that in another PR if it doesn't work

PR #279, opened 2021-01-11
updating autobump sciprt to call from piona

jonahjon:

Based off Leo's Update-maintainers script, trying to update Autobump to use Piona as the default bot.

### DIFF
--- a/config/jobs/autobump/autobump.yaml
+++ b/config/jobs/autobump/autobump.yaml
@@ -33,6 +33,8 @@ periodics:
         readOnly: true
       env:
     # autobump.sh args
+      - name: GH_PROXY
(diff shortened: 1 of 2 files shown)


maxgio92:
Thanks @jonahjon! Maybe we can also remove the part about Gerrit? WDYT?

jonahjon:
> Thanks @jonahjon! Maybe we can also remove the part about Gerrit? WDYT?

I don't mind either way, it's part of the original but also serves no purpose for us. I'll take it if you like though.

maxgio92:
Yes, no problem. Let's do it in a different PR

maxgio92 (approved):

PR #282, opened 2021-01-15
using alpine with git installed for terrascan

jonahjon:

Terrascan needed Git in image


maxgio92 (approved):
/lgtm

PR #292, opened 2021-01-22
Use Service Account authentication on update-jobs for in-cluster usage

maxgio92:
To rebase on master after #290 being merged.

Main refactoring work:
- Ship OCI image with compiled `update-jobs`
- Use Service Account token authentication with dedicated RBAC instead of system:masters privileges and authentication via kubeconfig
- Remove AWS CLI dependency on OCI image

/cc @jonahjon


prow/update-jobs/Makefile
```diff
@@ -1,36 +1,67 @@
...
+build: $(OUTPUT_PATH)/$(PROJECTNAME)
+	@echo "  >  Building binary... "
 	touch $(STDERR)
 	rm $(STDERR)
-	mkdir -p $(OUTPUT)
-	go build -o "$(OUTPUT)/$(PROJECTNAME)" $(GOFILES) 2> $(STDERR)
-	cat $(STDERR) | sed -e '1s/.*/\nError:\n/'  | sed 's/make\[.*/ /' | sed "/^/s/^/     /" 1>&2
+	mkdir -p $(OUTPUT_PATH)
+	go build -o "$(OUTPUT_PATH)/$(PROJECTNAME)" $(GOFILES)
+	@echo "Compiled in: $(OUTPUT_PATH)/$(PROJECTNAME)"
```
leodido, line 37: The `cat ...` here is needed to show the errors that occur during the build

leodido (commented):
Hey Max, there's a thing I'm not getting...

Shouldn't the logic of the `update-jobs` image (building, pushing, dockerfile) be inside the `image/update-jobs` directory?

This PR seems to move some logic from there inside the tool directory (ie., `prow/update-jobs`).

What's the motivation for this?

Can't we obtain the same 3 goals listed in the PR corpus by editing the Makefile in `image/update-jobs`?

prow/update-jobs/Makefile
maxgio92: Shouldn't we be able to read it?  Anyway I can revert it  :)

maxgio92 (commented):
> Hey Max, there's a thing I'm not getting...
> 
> Shouldn't the logic of the `update-jobs` image (building, pushing, dockerfile) be inside the `image/update-jobs` directory?
> 
> This PR seems to move some logic from there inside the tool directory (ie., `prow/update-jobs`).
> 
> What's the motivation for this?
> 

Hi @leodido ! 
Yes the Dockerfile and image related should, but at the same time we would build that image from building sources in order to include the compiled update-jobs, so I think sources and image source should be in the same folder. WDYT?

> Can't we obtain the same 3 goals listed in the PR corpus by editing the Makefile in `image/update-jobs`?

For goal 1) and 2) to authenticate update-jobs with service account in `main.go` we should get the client config via `rest.InClusterConfig()` to authenticate with the mounted service account token - at the same time we can support always authentication with certs by providing the kubeconfig.
In turn the Service Account should be in place with RBAC configured as in `config/prow/update-jobs.yaml`.
At the same time the `Dockerfile` I think no longer needs the AWS CLI for the same goal.

With goal 3) by shipping the compiled update-jobs with the image we can use it as the default command, (maybe also the entrypoint and passing `update-jobs` arguments as command in `config/jobs/update-jobs/update-jobs.yaml`).

What do you think? Maybe I'm missing something :)

prow/update-jobs/Makefile
maxgio92: In the meantime I reverted it

config/jobs/update-jobs/update-jobs.yaml
```diff
@@ -7,11 +7,11 @@ postsubmits:
...
       containers:
-      - command:
-        - /workspace/update-jobs.sh
-        env:
-        - name: AWS_REGION
-          value: eu-west-1
         image: 292999226676.dkr.ecr.eu-west-1.amazonaws.com/test-infra/update-jobs:latest
         imagePullPolicy: Always
+        command:
+          - /go/bin/update-josb
```
jonahjon, line 15: Typo on line 15 `- /go/bin/update-jobs`

prow/update-jobs/Makefile
jonahjon: Agree let's edit the existing Dockerfile in `images/update-jobs`

config/prow/update-jobs.yaml
```diff
@@ -28,4 +28,24 @@ roleRef:
...
\ No newline at end of file
+  name: system:nodes
+---
+kind: ClusterRoleBinding
+apiVersion: rbac.authorization.k8s.io/v1
+metadata:
+  name: update-jobs-sa
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
```
jonahjon, line 39: Should this Role instead of ClusterRole since we making it namespace scoped?

config/prow/update-jobs.yaml
```diff
@@ -28,4 +28,24 @@ roleRef:
...
+- apiGroup: rbac.authorization.k8s.io
+  kind: ServiceAccount
+  name: update-jobs
+  namespace: prow
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ServiceAccount
+metadata:
+  name: update-jobs
+  namespace: prow
```
jonahjon, line 51: the namespace we run pods in is `test-pods` and the namespace prow microservices run in `default` so this should be changed.

prow/update-jobs/Dockerfile
```diff
@@ -0,0 +1,35 @@
+# --- Args ---
```
jonahjon, line 1: Change `ARG` to `ENV` for any values being used in the image such as `GOPATH` which might be used in the entrypoint script.

config/prow/update-jobs.yaml
maxgio92: Yeah, Role to enable write configmaps on prow namespace (in our case default), with service account in test-pods

prow/update-jobs/Makefile
maxgio92: Agree @leodido @jonahjon I got the point, let's keep images and prow sources separate :)

prow/update-jobs/Dockerfile
maxgio92: Done. Align environment to goversion build argument

prow/update-jobs/Makefile
maxgio92: I noticed that also packages downloading info are written to stderr

maxgio92:
Local tests just passed (`make test` under images/update-jobs).
Update Jobs job it seems is now able to update Prow Jobs configurations with its `Service Account`.
Least privilege RBAC is applied in `config/prow/update-jobs.yaml`.
The related `ProwJob` configuration would pick the new pushed (`latest`) version, introduced by this PR.

What do you think?

jonahjon:
> Local tests just passed (`make test` under images/update-jobs).
> Update Jobs job it seems is now able to update Prow Jobs configurations with its `Service Account`.
> Least privilege RBAC is applied in `config/prow/update-jobs.yaml`.
> The related `ProwJob` configuration would pick the new pushed (`latest`) version, introduced by this PR.
> 
> What do you think?

Lets send it!

jonahjon (approved):

PR #296, opened 2021-01-25
removing GERRIT env's and script calls from autobump.sh

jonahjon:


leodido (approved):
LGTM

PR #297, opened 2021-01-26
fixing SA typo, and adding to periodic

jonahjon:

Looks like SA role had a typo in it, tested out with sleep pod, and uploads seemed to work now.


leodido (approved):
LGTM

PR #319, opened 2021-02-24
new build-libs job for falcosecurity/libs

leogr:
This PR aims to introduce an initial prow job for falcosecurity/libs CI, as per https://github.com/falcosecurity/libs/issues/9

If it works other similar jobs will be configured later.

Currently, just a draft.

### DIFF
--- a/config/jobs/build-prow-images/build-images.yaml
+++ b/config/jobs/build-prow-images/build-images.yaml
@@ -141,3 +141,31 @@ presubmits:
           privileged: true
       nodeSelector:
         Archtype: "x86"
+  - name: build-images-build-libs
+    decorate: true
+    path_alias: github.com/falcosecurity/test-infra
+    skip_report: false
+    agent: kubernetes
+    run_if_changed: '^images/update-libs/'
+    branches:
+      - ^master$
+    spec:
+      containers:
+      - command:
+          - "/home/prow/go/src/github.com/falcosecurity/test-infra/images/publish.sh"
+        args:
+          - "/home/prow/go/src/github.com/falcosecurity/test-infra/images/build-libs"
+        env:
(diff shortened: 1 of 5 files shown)


jonahjon (dismissed):

images/build-libs/Dockerfile
```diff
@@ -0,0 +1,28 @@
+FROM debian:stable
```
maxgio92, line 1: maybe we could pin a version?

leogr: I haven't pinned a version because there's not yet a clear requirement on falcosecurity/libs about that. I believe once we clean up and fine-tune the build system, it will become evident what image should be used and consequently an image version to be pinned.
Basically, just assume `debian:stable` as temporary solution :)

config/jobs/build-prow-images/build-images.yaml
```diff
@@ -139,5 +139,33 @@ presubmits:
...
         securityContext:
           privileged: true
+      nodeSelector:
+        Archtype: "x86"
+  - name: build-images-build-libs
+    decorate: true
+    path_alias: github.com/falcosecurity/test-infra
+    skip_report: false
+    agent: kubernetes
+    run_if_changed: '^images/update-libs/'
```
leogr, line 149:
```suggestion
    run_if_changed: '^images/build-libs/'
```
Wrong copy-paste

leogr (commented):
Just found a minor issue, I'm fixing that

images/build-libs/Dockerfile
maxgio92: Oh ok sure :) Thanks for the explanation

maxgio92 (approved):

PR #320, opened 2021-02-25
Add build-libs AWS ECR Repository

maxgio92:
An entry for creation of the `build-libs` OCI image repository on AWS ECR is added with this PR.


leodido (approved):

leogr:
Thank you! I missed that, good catch!

PR #336, opened 2021-03-18
Change the email @poiana uses for its operations

leodido:


images/autobump/autobump.sh
```diff
@@ -1,13 +1,14 @@
 #!/usr/bin/env bash
 
-# Copyright 2019 The Kubernetes Authors.
+#!/usr/bin/env bash
 #
```
leodido, line 4:
```suggestion
```

jonahjon (approved):

PR #339, opened 2021-03-22
Amazon Linux 2 5.4 kernels

zuc:
SSIA.


leodido (commented):
LGTM

/hold

Holding it because we first need to release a new version of driverkit

fntlnz (dismissed):

leodido (approved):

PR #349, opened 2021-03-26
Add jobs to push falco-exporter and falco-sidekick plain manifests from Helm charts

maxgio92:
This PR enable the image `update-falco-k8s-manifests` to `template` every Helm chart from `falcosecurity` Helm repository and push them to `evolution` repo.
It also adds two new jobs to generate and push manifests for `falco-exporter` and `falco-sidekick`.


leodido (approved):
LGTM

leogr (commented):
LGTM
Maybe we should have them start at different times?

maxgio92:
> LGTM
> 
> Maybe we should have them start at different times?

I don't it might be an issue to run in parallel, I'd expect they would require low amount of resources

PR #355, opened 2021-04-01
Generate Kustomizations for Falco ecosystem components

maxgio92:
This PR updates the `update-falco-k8s-manifests` `ProwJob` image to include a base `Kustomization` file for components of which Kubernetes manifests are generated, in order to provide a version-zero support for users that use Kustomize to manage Kubernetes configurations.

So, in the end for who wants to manage his configurations with Kustomize will (from [evolution](https://github.com/falcosecurity/evolution) root folder):
```
kubectl apply -k deploy/kubernetes/falco
```
and who doesn't:
```
kubectl apply -f deploy/kubernetes/falco/templates
```

### DIFF
--- a/images/update-falco-k8s-manifests/Dockerfile
+++ b/images/update-falco-k8s-manifests/Dockerfile
@@ -12,7 +12,8 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
     curl \
     && apt-get clean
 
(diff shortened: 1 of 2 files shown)


jonahjon (approved):
It looks fine, if you've tested this out let's do it

PR #374, opened 2021-04-21
feat Attempting to add a better regex for driver-kit jobs

jonahjon:

Currently all the build driver jobs trigger on a new driver addition.

This would attempt to build out a regex for each one to reduce the test scope.


config/jobs/build-drivers/build-drivers.yaml
```diff
@@ -6,7 +6,7 @@ presubmits:
     agent: kubernetes
     branches:
       - ^master$
-    run_if_changed: '^driverkit\/'
+    run_if_changed: '^driverkit/config/(2aa88dcf6243982697811df4c1b484bcbe9488a2|5c0b863ddade7a45568c0ac97d037422c9efb750|85c88952b018fdbce2464222c3303229f5bfcfad|ae104eb20ff0198a5dcb0c91cc36c86e7c3f25c7)/amazonlinux_.*'
```
leodido, line 9: Sounds like the slash needs a backslash? Am I missing something?

Ie., `\/`.

config/jobs/build-drivers/build-drivers.yaml
```diff
@@ -32,7 +32,7 @@ presubmits:
     agent: kubernetes
     branches:
       - ^master$
-    run_if_changed: '^driverkit\/'
+    run_if_changed: '^driverkit/config/(2aa88dcf6243982697811df4c1b484bcbe9488a2|5c0b863ddade7a45568c0ac97d037422c9efb750|85c88952b018fdbce2464222c3303229f5bfcfad|ae104eb20ff0198a5dcb0c91cc36c86e7c3f25c7)/amazonlinux2_.*'
```
leodido, line 35:
```suggestion
    run_if_changed: 
'^driverkit\/config\/(2aa88d|5c0b86|85c889|ae104e)[a-z0-9]{34}\/amazonlinux2_.*'
```

What do you think to shorten the regex to something like ^^^ ?

leodido: Does the regex engine in use support character classes (eg., `[a-z]`) and quantifiers (eg., `{34}`)?

leodido (commented):
I like these improvements!

Anyways, I left some comments.

config/jobs/build-drivers/build-drivers.yaml
jonahjon: The backslash I think is for a directory level regex, since this is triggering on files changes/additions I don't believe it is needed.

config/jobs/build-drivers/build-drivers.yaml
jonahjon: It does support that, I'll update the PR once I get it working. Using https://regoio.herokuapp.com/ to test

leodido (approved):

PR #380, opened 2021-04-27
Fix/terraform lifecycle rule

jonahjon:
`pr-logs` is where the bulk of the logs live. Terraform is having issues with 2 life_cycle rules changing the rule to only add the important one will allow the CircleCI job will pass again.


maxgio92:
Thanks @jonahjon, I'm going to check

jonahjon:
The build-driver got triggered by a rule update.... I added an issue for that bug but might have to re-open PR

PR #381, opened 2021-04-29
Configure Deck to get PR status from Github

### LINKED ISSUE #23: Setup PR status page
image

maxgio92:
This PR introduces the PR status page.

This is accomplished by configuring OAuth so that Deck can access GH resources via API, on behalf of the authenticated Github identity.

Closes #23.

### DIFF
--- a/tools/deploy_prow.sh
+++ b/tools/deploy_prow.sh
@@ -37,6 +37,10 @@ function launchConfig(){
   kubectl create secret generic hmac-token --from-literal=hmac="$(./tools/1password.sh -d config/prow/hmac-token)" || true
   kubectl create secret generic oauth-token --from-literal=oauth="$(./tools/1password.sh -d config/prow/oauth-token)" || true
   kubectl create secret generic oauth-token --from-literal=oauth="$(./tools/1password.sh -d config/prow/oauth-token)" -n test-pods || true
+
+  # PR Status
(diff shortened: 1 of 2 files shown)


tools/deploy_prow.sh
```diff
@@ -37,6 +37,10 @@ function launchConfig(){
   kubectl create secret generic hmac-token --from-literal=hmac="$(./tools/1password.sh -d config/prow/hmac-token)" || true
   kubectl create secret generic oauth-token --from-literal=oauth="$(./tools/1password.sh -d config/prow/oauth-token)" || true
   kubectl create secret generic oauth-token --from-literal=oauth="$(./tools/1password.sh -d config/prow/oauth-token)" -n test-pods || true
+
+  # PR Status
+  kubectl create secret generic github-oauth-config --from-literal=oauth="$(./tools/1password.sh -d prow-prstatus-github-oauth-app.yaml)" || true
```
leodido, line 42: Would it make sense to add this to the root [Makefile](https://github.com/falcosecurity/test-infra/blob/master/Makefile) too? WDYT?

jonahjon (approved):
Awesome, hope this works it will be really cool to have.

tools/deploy_prow.sh
maxgio92: Oh sure, that `Makefile` is to be updated with this

maxgio92: Also, I think that we should unify scripts with `Makefile`s without repeating

PR #383, opened 2021-05-03
fix(config/clusters/storage.tf): s3 bucket lifecycle aws provider syntax

maxgio92:
This PR should fix and unblock syntax errors that now are blocking the CI/CD of the infrastructure.
/cc @jonahjon WDYT?


jonahjon (approved):
Love it, thanks for helping on this... my last PR got bogged.

PR #401, opened 2021-05-22
Update driver configurations (2021-05-22)

leodido:
Prebuilt drivers update


maxgio92:
Hey @leodido what do you think about putting also these updates into a periodic prow job?

maxgio92 (approved):

Issue #406, opened 2021-05-25
Further split drivers build jobs


maxgio92:
The current version of drivers build jobs is split per Linux distribution (and per distro variant where present).
For some distro there a lot of driver versions to build and this causes very long running time of related jobs.

The proposed feature is to have a more fine-grained DBG jobs.
This could prevent from situations where issues like #379 prevent from merging PRs for days.

The other benefit is to better spread workload on distributed computing resources, enabling for better parallelization.

PR #410, opened 2021-05-26
Split drivers build jobs

maxgio92:
This PR aims to starting implement what is proposed in FR #406.

- First, the [`build-drivers`](https://github.com/falcosecurity/test-infra/tree/master/images/build-drivers) image script now accepts a optional and positional arguments as pattern string to also:
  - filter by kernel version (as expected by [DBG](https://github.com/falcosecurity/test-infra/tree/master/driverkit) [`specific_target`](https://github.com/falcosecurity/test-infra/tree/master/driverkit#specific-target) Make target)
  - filter by driver version
- Driver build jobs for `ubuntu-generic` and `ubuntu-aws` distro variants are split into jobs for kernel version 3, 4 and 5
- Furthermore, the `build-drivers.sh` script has been refactored, so that we can also call it like that (from the root of this repo):
```sh
PUBLISH_S3=false ENSURE_DOCKER=false DBG_WORKDIR=$PWD/driverkit \
./images/build-drivers/build-drivers.sh amazonlinux 4.14.77 85c88952b018fdbce2464222c3303229f5bfcfad
```


maxgio92:
What do you think @jonahjon?

jonahjon:
@maxgio92 I think this is awesome idea, lets give it a whirl and test it.

images/build-drivers/build-drivers.sh
```diff
@@ -27,6 +27,13 @@ function start_docker() {
 }
 
 PUBLISH_S3="${PUBLISH_S3:-false}"
+
+TARGET_DISTRO="${1}"
+TARGET_KERNEL="${2}"
+DBG_ENV_PARAMS="-e TARGET_DISTRO=${TARGET_DISTRO}"
```
leodido, line 33: There's any reason to do not check the `TARGET_DISTRO` value?

images/build-drivers/build-drivers.sh
```diff
@@ -46,9 +53,11 @@ echo "******************************************************"
 start_docker
 
 cd driverkit/
-make -e TARGET_DISTRO="$1" specific_target
 
-test "${PUBLISH_S3}" == "true" && make publish_s3
+make $DBG_ENV_PARAMS specific_target
```
leodido, line 57: Finally, seeing you Max using my beloved Makefiles makes me so happy! 🤗

leodido (commented):
This is lovely Max! Finally 👏 

Left one comment :)

Also: you probably forgot to create the ProwJobs for `ubuntu-generic` 3.* kernels (we have a bunch of them) ;)

images/build-drivers/build-drivers.sh
maxgio92: 😄 Make is so powerful, why not using them!? Then, they're well-written I think!

images/build-drivers/build-drivers.sh
maxgio92: I don't think so. I wrongly leaved it with the assumption that is always passed. Going to fix it

maxgio92:
I updated the description since I added a refactoring on the `build-drivers.sh` script

images/build-drivers/build-drivers.sh
maxgio92: I did a refactoring @leodido on `build-drivers.sh` script. So check each argument in a way that we can also accept a third to specify the `TARGET_VERSION` (the version of the Falco driver). Let me know WDYT, so that in case I resolve this review.

In the end the script can now be called like that:
```sh
PUBLISH_S3=false ENSURE_DOCKER=false DBG_WORKDIR=$PWD/driverkit \
  ./images/build-drivers/build-drivers.sh amazonlinux 4.14.77 85c88952b018fdbce2464222c3303229f5bfcfad
```

maxgio92:
The local tests passed.

images/build-drivers/build-drivers.sh
maxgio92: (Also, added the jobs for Linux kernel 3.*)

maxgio92:
At this point, as of #421, if other `build-drivers` work, this PR can be considered already working and approved :D

leodido (approved):

PR #432, opened 2021-06-15
Starting cluster upgrades

jonahjon:

Writing instructions, and starting process of updating cluster components.

More PR's will come out for the other upgrades after, starting with cluster version since it is a requirement for some of the other ones due to Kube API versions.


maxgio92 (approved):
Thanks @jonahjon to start documenting the component versions, great idea!

/approve

PR #433, opened 2021-06-16
fixing terrascan, and tightning SG rules

jonahjon:

Another Terrascan addition. We can't really scope these SG's to any /32 since we don't use a bastion host at the moment. Would rather install SSM agent daemon and remove this in the future for SSH access.

When reviewing the rules noticed they were scoped for other network ranges not in use so fixed that.


maxgio92:
These rules changes are changing too often than how much we can cope with :D
Sometimes not so relevant IMHO... :/

maxgio92 (approved):
/approve

Issue #451, opened 2021-06-25
Build Drivers False Positive


jonahjon:
**Describe the bug**

Prow build drivers job is reporting healthy on build failures

https://prow.falco.org/view/s3/falco-prow-logs/pr-logs/pull/falcosecurity_test-infra/448/build-drivers-centos-presubmit/1408049337874780160

This job from PR https://github.com/falcosecurity/test-infra/pull/448 marked as passed even though the job failed.

This leads me to believe this line in the build driver script isn't passing error up the main function since it's where the failure occured.

https://github.com/falcosecurity/test-infra/blob/master/images/build-drivers/build-drivers.sh#L76

We need to make this function pass error back to prow otherwise it just finishes the `main` function with return of `0`
 
https://github.com/falcosecurity/test-infra/blob/master/images/build-drivers/build-drivers.sh#L59

sachinkumarsingh092:
@jonahjon if there is no issue with `patsubst` in the Makefile for the target and the substitutions are working fine, can a simple `||` condition for `exit 1` on [this line](https://github.com/falcosecurity/test-infra/blob/master/images/build-drivers/build-drivers.sh#L76) work?

maxgio92:
I think the important motivation is that we're building drivers with a best effort approach.
There always are drivers that are not pre-built, so we potentially have driver build jobs that constantly fail.
As that said, we must (might?) go on and build what we can.

Anyway, we can reason about that, but I don't think is a matter of changing the exit code. Otherwise we won't be able to build and publish as for now, IMHO.

WDYT?

PR #460, opened 2021-06-30
WIP Build drivers bug fix

jonahjon:
There is currently a bug for the build drivers that will have a job finish as "success" even when there is a failure. Luckily the publish job doesn't trigger, but this doesn't allow the re-testing mechanism we would expect.

https://github.com/falcosecurity/test-infra/issues/451

Trying to adjust the makefile and the line where it fails to actually emit a trap failure, and adding a fake driverkit entry to cause a failure. 

Previous
`@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 || echo $(3) >> output/failing.log`

Adding the `&&` so it will exit on failures.
`@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 && { echo $(3) >> output/failing.log; exit 1; }`

### DIFF
--- a/driverkit/Makefile
+++ b/driverkit/Makefile
@@ -31,7 +31,7 @@ define gen_build_targets
 $(1): $(3)
 	utils/checkfiles $(3)
 	@mkdir -p output/$(2)
-	@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 && echo $(3) >> output/failing.log
+	@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 && { echo $(3) >> output/failing.log; exit 1; }
 endef
 $(foreach CONFIG,$(CONFIGS),\
 	$(eval INNER := $(patsubst config/%,%,$(CONFIG)))\
--- a/driverkit/Makefile
+++ b/driverkit/Makefile
@@ -31,7 +31,7 @@ define gen_build_targets
 $(1): $(3)
 	utils/checkfiles $(3)
 	@mkdir -p output/$(2)
-	@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 && { echo $(3) >> output/failing.log; exit 1; }
+	@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 || { echo $(3) >> output/failing.log; exit 1; }
 endef
 $(foreach CONFIG,$(CONFIGS),\
 	$(eval INNER := $(patsubst config/%,%,$(CONFIG)))\
--- a/driverkit/Makefile
+++ b/driverkit/Makefile
@@ -31,7 +31,7 @@ define gen_build_targets
 $(1): $(3)
 	utils/checkfiles $(3)
 	@mkdir -p output/$(2)
-	@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 || echo $(3) >> output/failing.log
+	@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 && echo $(3) >> output/failing.log
 endef
 $(foreach CONFIG,$(CONFIGS),\
 	$(eval INNER := $(patsubst config/%,%,$(CONFIG)))\


maxgio92:
Thanks @jonahjon this is a good catch I think, as we discussed.
I leave some comments.

driverkit/Makefile
```diff
@@ -31,7 +31,7 @@ define gen_build_targets
 $(1): $(3)
 	utils/checkfiles $(3)
 	@mkdir -p output/$(2)
-	@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 || echo $(3) >> output/failing.log
+	@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 && echo $(3) >> output/failing.log
```
maxgio92, line 34: I only have a couple of concerns:
- we're fulfilling the `failing` log with `successful` builds
- as drivers build can be time-expensive, we could have a time-expensive trial & error if we exit on first failure
- I'd keep DBG and build jobs separate as DBG can be used also locally (for example maybe I'd want to build locally all the drivers and see only at the end the result avoiding trial and error)

My proposals below.

#### Control inside DBG

if we want to exit on first failure, in `driverkit/Makefile`:

```
@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 || { echo $(3) >> output/failing.log && exit 1 ; }
```
#### Control inside build job

If we don't change the DBG and check after DBG run complete, in `images/build-drivers/build-drivers.sh`:

```
$make $DBG_MAKE_BUILD_OPTIONS $DBG_MAKE_BUILD_TARGET

test -s output/failing.log || { pretty_echo "DBG build failed"  && cat output/failing.log && exit 1 ; }

pretty_echo "DBG build complete"
```

maxgio92 (approved):

driverkit/Makefile
leodido: I don't think this is a viable solution.

That `or` operator was meant to make the Makefile write the failing builds to `output/failing.log`

If the problem is making it exit with status != 0 in that case .. Well just use `exit 1` on that conditional branch :)

leodido:
```suggestion
	@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 && { echo $(3) >> output/failing.log; exit 1; }
```

leodido (commented):
Please see my suggestion/comments :)

driverkit/Makefile
maxgio92: Maybe the first `&&` should remain an or `||`, right @leodido?

driverkit/Makefile
```diff
@@ -31,7 +31,7 @@ define gen_build_targets
 $(1): $(3)
 	utils/checkfiles $(3)
 	@mkdir -p output/$(2)
-	@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 || echo $(3) >> output/failing.log
+	@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 && { echo $(3) >> output/failing.log; exit 1; }
```
maxgio92, line 34:
```suggestion
@${DRIVERKIT} docker -c $(3) --driverversion $(2) --timeout 1000 || { echo $(3) >> output/failing.log; exit 1; }
```
so that we exit only when failing build

maxgio92 (changes requested):

jonahjon:
This seemed to pass, can we simulate a driver build failure?

maxgio92:
@jonahjon this should fail on the 3th (out of 7) config (`config/17f5df52a7d9ed6bb12d3b1768460def8439936d/debian_5.9.0-1-amd64_1.yaml`):
```
make -e TARGET_DISTRO="debian" -e TARGET_KERNEL="5.9.0" -e TARGET_VERSION=17f5df52a7d9ed6bb12d3b1768460def8439936d specific_target
```

My only concern is about how much of them are already failing and we prevent from publishing the successing ones?

maxgio92 (approved):
/approve

jonahjon:
Building drivers is optimistic, and letting this change go through could block other development. Talked with Max, and should think a bit more about this before merging it.

jonahjon:
Closing this for now.

PR #483, opened 2021-08-08
updating cluster autoscaler

jonahjon:

Cluster Autoscaler was older, and has throttling issues leading to crashes.... updating to new version.


maxgio92 (approved):
/approve

PR #479, opened 2021-08-03
amazonlinux2 5.4.105-48.177 support for EKS worker

guraslan:

PR #493, opened 2021-08-17
creating signed PR for #479 to add EKS AMI support

jonahjon:

This is exact same as PR #479 just a signed off PR so we can close that one out.


leogr (approved):
:+1:

PR #515, opened 2021-09-20
Add plugin sdk go jobs

mstemm:
This adds prow/poiana support for falcosecurity/plugin-sdk-go.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -211,6 +211,10 @@ branch-protection:
               protect: true
         plugins:
           branches:
+            main:
+              protect: true
+        plugin-sdk-go:
+          branches:
             master:
               protect: true
         test-infra:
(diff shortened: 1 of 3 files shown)


config/config.yaml
```diff
@@ -209,6 +209,12 @@ branch-protection:
           branches:
             master:
               protect: true
+        plugin-sdk-go:
+          required_pull_request_reviews:
+            required_approving_review_count: 2
```
leogr, line 214:
```suggestion
```

Since it's an incubating project, we could stick with only 1 approval at the moment.

config/config.yaml
```diff
@@ -209,6 +209,10 @@ branch-protection:
           branches:
             master:
               protect: true
+        plugin-sdk-go:
+          branches:
+            master:
```
leogr, line 214:
```suggestion
            main:
```
I have noticed right now that the plugin-sdk-go default branch is `main` (and not master).

leogr (commented):
Hey @mstemm 
I have joust found another small issue

config/config.yaml
mstemm: Ok fixed.

jonahjon (approved):

PR #516, opened 2021-09-20
Add plugins jobs

mstemm:
Add support for falcosecurity/plugins repo in prow/poiana.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -209,7 +209,7 @@ branch-protection:
           branches:
             master:
               protect: true
-        template-repository:
+        plugins:
           branches:
             master:
               protect: true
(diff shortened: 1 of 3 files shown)


config/config.yaml
```diff
@@ -209,7 +209,9 @@ branch-protection:
           branches:
             master:
               protect: true
-        template-repository:
+        plugins:
+          required_pull_request_reviews:
+            required_approving_review_count: 2
```
leogr, line 214:
```suggestion
```

Since it's an incubating project, we could stick with only 1 approval at the moment.

mstemm: Ok, fixed.

leogr: Thank you!

leogr (approved):
/approve

PR #517, opened 2021-09-21
Fix build-plugins/build-plugin-sdk-go jobs

mstemm:
Instead of using the local golang image, use golang:1.15. This matches
what's in go.mod for the plugins.

Also fix the command to use the the mount location for the code, as
a workingDir is set for the k8s job.

Signed-off-by: Mark Stemm <mark.stemm@gmail.com>


leogr (approved):
/approve

PR #520, opened 2021-09-25
Add driverkit config files for driver version 5727c45

zuc:
This PR:
1. adds driverkit config files for https://github.com/falcosecurity/libs/commit/5727c456ce22f3c1da8c3b1d7d6b6937a9b2126b since [falco 0.30.0](https://github.com/falcosecurity/falco/milestone/19) (due in ~5 days from now) depends on that
2. removes driverkit config files related to old driver versions as per the [artifacts cleanup proposal](https://github.com/falcosecurity/falco/blob/master/proposals/20200901-artifacts-cleanup.md#drivers)

### Notes for the reviewer(s)
The above mentioned cleanup proposal was written at a time in which Bintray hosting (and its limited storage capacity) was still a thing. Since we then moved to AWS S3, this cleanup pattern seems now unnecessary to me, but we don't have an updated version of that proposal yet. **Please advise on a suggested behaviour here (should we keep cleaning up or should we stop and amend the proposal?)**

Also, someone with the appropriate permissions should run [this Make target](https://github.com/falcosecurity/test-infra/blob/master/driverkit/Makefile#L24) to actually remove deprecated driver versions from S3 once this PR gets merged in.


leogr (approved):
/approve

LGTM, just a note for the future.


Although this PR is removing old prebuilt configurations (according to our process, we have to maintain just the last 3 driver versions), all old pre-builts will still be available on S3 for download (because of the cleanup job is not active). Doing that has the benefit that driver building process will not take so long as before (so far, it has been a pain). 
I believe this is needed, especially now since the Falco release should happen within two days. We would risk blocking the release otherwise. Eventually, we could decide later what to do with the old prebuilts (anyway, I think it's good to build only the last 3 versions but still leave previous ones on S3). 
So, before taking any further action, this topic needs to be discussed with core maintainers.

leodido:
I agree.

The proposal already mentioned S3 ([here](https://github.com/falcosecurity/falco/blob/master/proposals/20200901-artifacts-cleanup.md#archivation)). I'll take a stab at updating it deprecating the part about the cleanup job (that is no longer in place for the reasons you have correctly explained @leogr ).

leodido:
There you go: https://github.com/falcosecurity/falco/pull/1738

PR #536, opened 2021-10-12
Create debian_4.19.0-18

movd:
Signed-off-by: Moritz von Doetinchem <moritz.doetinchem@dg-i.net>


leogr:
Hey @movd 

Could you add the same configurations for the other driver versions too?
Thanks in advance!

PR #555, opened 2021-11-12
update(build-libs): test minimal libs build

FedeDP:
Note that BPF build is disabled because MINIMAL_BUILD does not support building bpf probe.


leogr (approved):
/approve

PR #595, opened 2022-01-14
update driverkit

Issif:
We just released a new version of driverkit which fixes issues we have for debian kernels > 5.x. This PR update the according docker image for using this new release.

https://github.com/falcosecurity/driverkit/releases/tag/v0.5.0

Signed-off-by: Issif <issif+github@gadz.org>

PR #596, opened 2022-01-15
adding ecr login to Makefile for base image pull

jonahjon:

Trying to fix https://github.com/falcosecurity/test-infra/pull/595

Most of our other builds use base images such as `golang` or `debian` so we didn't notice a lack of ECR login in the makefiles. Since this one uses a custom built image we needed an ECR login prior to build.


maxgio92 (approved):
/approve

PR #600, opened 2022-01-18
Add prow configs for falcosecurity/plugin-sdk-cpp

FedeDP:
**NOTE: no jobs added as it is an header only library thus there is no need to build it.**

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -648,7 +648,6 @@ tide:
         - do-not-merge/hold
         - do-not-merge/invalid-owners-file
         - do-not-merge/work-in-progress
-        - do-not-merge/release-note-label-needed
         - needs-rebase
       reviewApprovedRequired: true  
     - repos:
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -219,6 +219,10 @@ branch-protection:
           branches:
             master:
               protect: true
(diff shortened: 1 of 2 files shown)


config/plugins.yaml
```diff
@@ -848,6 +858,30 @@ plugins:
...
+    - dco
+    - dog
+    - goose
+    - help
+    - hold
+    - label
+    - lifecycle
+    - lgtm
+    - mergecommitblocker
+    - release-note
```
leogr, line 877:
```suggestion
```
We probably do not want the release note block, since the repo is not so big and it would be cumbersome.

config/config.yaml
```diff
@@ -632,6 +637,20 @@ tide:
...
+      labels:
+        - approved
+        - lgtm
+        - "dco-signoff: yes"
+      missingLabels:
+        - do-not-merge
+        - do-not-merge/hold
+        - do-not-merge/invalid-owners-file
+        - do-not-merge/work-in-progress
+        - do-not-merge/release-note-label-needed
```
leogr, line 651:
```suggestion
```

FedeDP:
Thanks @leogr !

leogr (approved):
/approve

PR #601, opened 2022-01-18
update: fix plugin name regex in build-plugins jobs

jasondellaluce:
Signed-off-by: Jason Dellaluce <jasondellaluce@gmail.com>

This updates the `build-plugins` job to fix the regex used to match plugin names.


leogr (approved):
perfect thank you!

/approve

PR #524, opened 2021-10-01
Avoid re-building all drivers each time

zuc:
This PR aims at reducing drivers building time.
More in detail:
- adds a `SKIP_EXISTING` (default: `true`) env var inside driverkit Makefile
- wraps driver building steps in a separate bash util that checks for S3 existence prior to building
- replaces `build-drivers` Prow jobs with `build-new-drivers` ones
- adds `build-all-drivers` Prow jobs

The rationale behind last two steps is as follows:
- In common scenarios, pre/post-submit jobs will build new drivers only (therefore skipping already existing ones and making the whole release of a new driver version sensibly shorter)
- If for any reason a complete build is needed in a timely fashion, `build-all-drivers` jobs can be manually invoked using @poiana's slash commands (e.g. `/test rebuild-all-drivers-amazonlinux2`) inside any PR targeting the main branch
- In any way, new full builds will be launched by periodic jobs the same way we use to do today 

Signed-off-by: Michele Zuccala <michele@zuccala.com>

PR #614, opened 2022-01-27
Fix bash condition that prevented SKIP_EXISTING from actually being honored

zuc:
This fixes a stupid bug that I introduced with #524.
Basically `SKIP_EXISTING` wasn't actually honored because *There Are No Booleans In Bash™*

/cc @leogr @maxgio92 


maxgio92:
Thanks @zuc!

maxgio92 (approved):
/approve

PR #615, opened 2022-01-28
Add amazonlinux2_4.14.246-187.474.amzn2.x86_64 BPF

mmoyerfigma:
This kernel already had a module defined, but no probe artifact. It compiles and runs the eBPF probe locally just fine, as far as I can tell.


maxgio92 (approved):
/approve

PR #621, opened 2022-02-11
Adding latest ebpf probe building to all amazonlinux2 kernels

abroglesc:
We have a lot of variety in our AWS environment and to cover our bases would like to build the eBPF probe for all amazonlinux2 kernels that we currently build the kernel module for. We have chosen to only do this for the latest driver used in `0.31.0`, `319368f1ad778691164d33d59945e00c5752cd27`, but can do all driver versions if this is desired.


abroglesc:
Check prow config failed and not sure how to restart that 😓

jonahjon (approved):

PR #348, opened 2021-03-26
[WIP] Prow jobs to compile falco for aarch64

zuc:
Signed-off-by: Michele Zuccala <michele@zuccala.com>

PR #626, opened 2022-02-18
update(images): improve autobump script regex to also support '-arm64' suffix

FedeDP:
`Signed-off-by: Federico Di Pierro <nierro92@gmail.com>`

The new `sed` will be:
```
's/gcr.io\/k8s-prow\/\([[:alnum:]_-]\+\):v[a-f0-9]\+-[a-f0-9]\+/gcr.io\/k8s-prow\/\1:v20220216-1007250f01/I'
```
where `v20220216-1007250f01` is the new version fetched by k8s test-infra.  

Examples:
stdin:
```
gcr.io/k8s-prow/deck:v20210216-1007250f01
gcr.io/k8s-prow/deck:v20120216-1007250f01-arm64
```
stdout:
```
gcr.io/k8s-prow/deck:v20220216-1007250f01
gcr.io/k8s-prow/deck:v20220216-1007250f01-arm64
```


maxgio92:
Thanks @FedeDP! I'd like to understand the context.
Considering that the test-infra is all running on x86 archs, is there a motivation for which we would want to consider also images for ARM?

Also, AFAIK multi-arch images should be able to keep a unique tag and the digest not to change, are there images as far as you know that instead are distribuited separately for each architecure?
For example I took a look here at the GCR of Deck image: https://console.cloud.google.com/gcr/images/k8s-prow/GLOBAL/deck

Thank you :-)

zuc:
Ciao @maxgio92,
actually we do already have some aarch64 nodes currently being provisioned [here](https://github.com/falcosecurity/test-infra/blob/master/config/clusters/eks.tf#L51-L69).

This PR is part of some work @FedeDP and I have been doing recently to start compiling falco for ARM too - basically taking over my initial work [here](https://github.com/falcosecurity/test-infra/pull/348).

There will be some more contribs coming in the next days/weeks on this, since there are still a few moving parts that need rethinking 🙂

zuc:
Also, pod utils are still not provided in a multiarch fashion from k8s folks, see https://console.cloud.google.com/gcr/images/k8s-prow/global/initupload as a reference 🤷‍♂️

maxgio92:
Hey @zuc, thank you for the context, got it. And thank you @FedeDP for this!

PR #637, opened 2022-02-23
fix(images): add libprotobuf-dev to deps in build-libs image

FedeDP:


leogr (approved):
/approve

PR #639, opened 2022-03-03
new(images,jobs): attempt to enable build-libs job on arm too

FedeDP:
* Adapted build-libs image makefile to overwrite ARCH argument for arm builds
* Added jobs to build and publish build-libs image from an arm node
* Added a build-libs-arm job to test build-libs on arm node

### DIFF
--- a/config/jobs/build-libs/build-libs.yaml
+++ b/config/jobs/build-libs/build-libs.yaml
@@ -23,6 +23,29 @@ presubmits:
             cpu: 1.5
       nodeSelector:
         Archtype: "x86"
+  - name: build-libs-arm
+    decorate: true
+    skip_report: false
+    agent: kubernetes
+    always_run: true  # Run for every PR, but master
+    spec:
+      containers:
+      - command:
+        - /build.sh
+        args:
+        - -DBUILD_BPF=On
+        - -DUSE_BUNDLED_DEPS=False
+        env:
+        - name: AWS_REGION
+          value: eu-west-1
+        image: 292999226676.dkr.ecr.eu-west-1.amazonaws.com/test-infra/build-libs:latest
+        imagePullPolicy: Always
+        resources:
+          requests:
+            memory: 3Gi
+            cpu: 1.5
+      nodeSelector:
+        Archtype: "arm"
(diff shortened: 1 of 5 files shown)


FedeDP:
I was able to test the new build-libs image on arm:
```
[==========] 39 tests from 6 test cases ran. (9 ms total)
[  PASSED  ] 39 tests.
[100%] Built target run-unit-test-libsinsp
Built target run-unit-tests
root@5072d3c07ec2:/workspace/libs/build#
```

I cloned the test-infra repo, checkout this PR branch, then:
`sudo ./images/build.sh /home/ec2-user/test-infra/images/build-libs`
to build the `build-libs` image for arm64, then `docker run` the image and manually run the build steps.

FedeDP:
I failed to test that the `build-images-build-libs-arm` job is actually working using the `tools/local_prowjob.sh` because 2 required images (https://github.com/falcosecurity/test-infra/blob/master/tools/local_prowjob.sh#L42) are not available for arm64.

So, assuming that the `build-images-build-libs-arm` job is correctly working, everything else is ok and tested.

FedeDP:
I am currently evaluating the usage of `docker buildx` instead; it would allow us to build and publish both amd64 and arm64 images from a single x86 node, without the additional `build-images-build-libs-arm` job.
Moreover, `buildx` would implicitly set `DOCKER_BUILDKIT=1`, allowing us to take the `TARGETARCH` arg inside the dockerfile, automatically set by docker itself (thus dropping the `ARG ARCH` line and the ugly `if` inside the makefile).

FedeDP:
So, our docker does not ship `buildx` subcmd unfortunately :(

FedeDP:
I think we'd need to bump docker-dind image docker version to support buildx.
@leogr @zuc WDYT? Is this worth it?

FedeDP:
> #9 51.28 Error while loading /usr/sbin/dpkg-split: No such file or directory
#9 51.28 Error while loading /usr/sbin/dpkg-deb: No such file or directory

(on arm64). That's interesting. I did not hit this while testing on arm64 machine.
Will double check!

zuc (approved):
/approve

Issue #405, opened 2021-05-25
Declarative Github teams setup


maxgio92:
As @leodido proposed we should have a way to declaratively manage Github Teams (maybe with Prow), so that we can manage the truth about access control for maintainers declaratively in Git.

leodido:
Let me expand a bit on this idea that I proposed in the falco-open-infra-maintainers.

## Problem statement

The OWNERS file declares the pool of maintainers from which @poiana chooses to ask for PR approvals and reviews for repository X.

@poiana requires that every maintainer in the OWNERS file is in the GitHub organization, but doesn't check the GitHub permissions of those maintainers.

A maintainer M needs to have GitHub "Maintain" privileges for a repository X in order to maintain it.

There is **not** an automatic and safe way to ensure that maintainer M is in the "X-maintainers" GitHub team (and not in any other), at the moment. 

The process is **manual*, error-prone, and very difficult to maintain itself. The process is **not** transparent.

## Current status

One of the GitHub organization admins create **manually** a GitHub team (conventional name: "X-maintainers") for every repository X.

Then, it adds to it the GitHub handles in the "approvers" section of the root OWNERS file of repository X.

## Temporary workaround

I propose to create a ProwJob that:

1. reads the OWNERS files (using [maintainers-generator](https://github.com/leodido/maintainers-generator) as I did for the `update-maintainers` ProwJob in the past) for every (non-archived) repository in the GitHub organization
2. obtains the people list (by querying GitHub APIs) for every GitHub team "<repo>-maintainers"
3. alerts on any GitHub team that should be missing
4. checks the set of approvers against the people list erroring out if the diff is not empty

Running this ProwJob periodically should help us spot manual errors.

## Long-term solution

Setup the GitHub teams automatically (similarly to what Kubernetes does into [org](https://github.com/kubernetes/org)).

We could use the [peribolos](https://github.com/kubernetes/test-infra/blob/master/prow/cmd/peribolos) tool for achieving this goal. In its simplest form, we'll have a new `orgs` key in the main Prow `config.yaml`.

## Resources

This [KubeCon talk](https://www.youtube.com/watch?v=te3Xj2zr1Co&ab_channel=CNCF%5BCloudNativeComputingFoundation%5D) may help.

maxgio92:
Thank you @leodido for the detailed explanation.

Maybe is worth going with the long-term solution? I don't think the effort for having a detection cronjob is worth we'll still need prevention.

For the long-term solution I think:

[...]

PR #643, opened 2022-03-04
Add peribolos to manage our Github org as code

zuc:
This PR addresses long time standing issue #405.

Note that `config/org.yaml` has been dumped directly from Github API on Fri Mar 4th 2022 at 5pm CEST.
Depending on when this get approved, we might need to resync it to embed possible updates happened in the meantime.

/cc @maxgio92 
/cc @jonahjon 
/cc @leogr 

### DIFF
--- a/config/jobs/peribolos/peribolos.yaml
+++ b/config/jobs/peribolos/peribolos.yaml
@@ -1,0 +1,108 @@
+presubmits:
+  falcosecurity/test-infra:
+    - name: peribolos-pre-submit
+      branches:
+        - ^master$
+      decorate: true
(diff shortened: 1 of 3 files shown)


maxgio92:
Do you think we could test this @zuc , before approving it?

zuc:
> Do you think we could test this @zuc , before approving it?

@maxgio92 Well you can easily see the result of applying this config file by manually running the container entrypoint with same params but omitting `--confirm` (this launches peribolos in dry-run mode).

Let me know if you need any help in doing that! 😊

maxgio92 (approved):
/approve

PR #648, opened 2022-03-08
Fix peribolos binary path and triggering conditions

zuc:
/cc @maxgio92 
/cc @leogr 


leogr:
I'm overriding the required check "peribolos-pre-submit" so that this PR can fix it.

PR #649, opened 2022-03-08
update(images): do not bind to old docker version for docker-dind image

FedeDP:

### DIFF
--- a/config/jobs/build-prow-images/build-images.yaml
+++ b/config/jobs/build-prow-images/build-images.yaml
@@ -223,5 +223,32 @@ presubmits:
         securityContext:
           privileged: true
       nodeSelector:
-        Archtype: "x86" 
-      
+        Archtype: "x86"
+  - name: build-images-build-docker-dind
+    decorate: true
+    path_alias: github.com/falcosecurity/test-infra
+    skip_report: false
+    agent: kubernetes
+    run_if_changed: '^images/docker-dind/'
+    branches:
+      - ^master$
+    spec:
+      containers:
+      - command:
+          - "/home/prow/go/src/github.com/falcosecurity/test-infra/images/build.sh"
+        args:
+          - "/home/prow/go/src/github.com/falcosecurity/test-infra/images/docker-dind"
+        env:
+        - name: AWS_REGION
+          value: eu-west-1
+        image: 292999226676.dkr.ecr.eu-west-1.amazonaws.com/test-infra/docker-dind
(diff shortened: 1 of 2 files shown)


FedeDP:
This allows us to make use of `docker buildx` for https://github.com/falcosecurity/test-infra/pull/639.

FedeDP:
Also, there is no job config for docker-dind image update.
How is it updated?

zuc:
I think we have been updating it manually (see https://github.com/falcosecurity/test-infra/blob/master/tools/push_all_images.sh#L6)
Maybe @jonahjon you know more about this?

zuc:
Also, @FedeDP would you mind creating a proper update workflow for `docker-dind` as part of your effort on this? 🙏

images/docker-dind/Dockerfile
```diff
@@ -27,10 +27,8 @@ RUN curl -fsSL https://download.docker.com/linux/$(. /etc/os-release; echo "$ID"
     "deb [arch=amd64] https://download.docker.com/linux/$(. /etc/os-release; echo "$ID") \
     $(lsb_release -cs) stable"
 
-ENV DOCKER_VERSION=18.06.1*
-
 RUN apt-get update && \
-    apt-get install -y --no-install-recommends docker-ce=${DOCKER_VERSION} && \
+    apt-get install -y --no-install-recommends docker-ce && \
```
zuc, line 31: I would prefer to keep the docker version explicit so that we can 1) track down changes and 2) easily attach an update workflow for this image using Prow jobs.

zuc (changes requested):

images/docker-dind/Dockerfile
maxgio92: I agree with @zuc

maxgio92:
@FedeDP as @zuc said there's for now no continuos integration for those images. Also because, unfortunately, there's no real versioning of the artifacts (`latest` tag only + `Always` imagePullPolicy).

maxgio92:
But @FedeDP good point IMHO, we would need that

images/docker-dind/Dockerfile
FedeDP: Makes sense!

FedeDP:
I added 2 new jobs to build and publish docker-dind image, plus manually bumped DOCKER_VERSION env variable in docker-dind Dockerfile.

zuc (approved):

PR #657, opened 2022-03-17
build ko 5.4.181-99.354.amzn2.x86_64

Kaizhe:
Signed-off-by: Kaizhe Huang <khuang@aurora.tech>

PR #665, opened 2022-03-24
adding new configuration files for 5.4.181-99.354.amzn2.x86_64

tchanos:
Signed-off-by: Timothy Chan <timothy.chan@outsystems.com>


zuc (changes requested):
@tchanos Thanks for your contribution.

I see you are forcing `kernelversion: 64` in order to obtain a `.x86_64` suffix in the driver name. Can you please explain why?

Also, it seems that support for AL2 5.4.181-99.354 has already been added via #657.
Why would we need this additional PR for the same kernel version in your opinion?

Thanks

tchanos:
> @tchanos Thanks for your contribution.
> 
> I see you are forcing `kernelversion: 64` in order to obtain a `.x86_64` suffix in the driver name. Can you please explain why?
> 
> Also, it seems that support for AL2 5.4.181-99.354 has already been added via #657. Why would we need this additional PR for the same kernel version in your opinion?
> 
> Thanks

Hi @zuc , thank you for the review... newbie error: I followed [PR656](https://github.com/falcosecurity/test-infra/pull/656/files) but now I realize maybe it is incorrect. I actually thought that it made sense, but I'll be following 657, and set this to 1 instead of 64.

About the AL2 5.4.181-99.354 added in #657, yes agreed, however it builds the Module and does not build the Probe (probe: output/319368f1ad778691164d33d59945e00c5752cd27/falco_amazonlinux2_5.4.181-99.354.amzn2.x86_64.o). That is the file that is attempted by the falco pod to download but fails. Is the Probe not needed, then not sure why the pod would want to download the `.o` file when only the module `.ko` exists.

Will work on pushing another commit with the kernel version set to 1.

zuc:
> About the AL2 5.4.181-99.354 added in #657, yes agreed, however it builds the Module and does not build the Probe (probe: output/319368f1ad778691164d33d59945e00c5752cd27/falco_amazonlinux2_5.4.181-99.354.amzn2.x86_64.o). That is the file that is attempted by the falco pod to download but fails. Is the Probe not needed, then not sure why the pod would want to download the `.o` file when only the module `.ko` exists.

In that case, you just need to add a `probe` key under the `output` dict in the very same files added by #657 - there’s no need to add new driverkit configurations.

As a reference, please see https://github.com/falcosecurity/test-infra/blob/master/driverkit/config/17f5df52a7d9ed6bb12d3b1768460def8439936d/amazonlinux2_5.4.176-91.338.amzn2.x86_64_1.yaml

zuc (approved):
/approve

PR #684, opened 2022-04-27
Add job to generate configs from crawled kernels

maxgio92:
Signed-off-by: Massimiliano Giovagnoli <me@maxgio.it>

/cc @FedeDP

### DIFF
--- a/config/jobs/update-dbg/update-dbg.yaml
+++ b/config/jobs/update-dbg/update-dbg.yaml
@@ -1,6 +1,6 @@
 periodics:
   - name: update-dbg
-    cron: "0 10 * * 6" # Run every Sunday at 10:00
+    cron: "0 10 5,20 * *"
     decorate: true
     extra_refs:
     # Check out the falcosecurity/test-infra repo
--- a/config/jobs/update-dbg/update-dbg.yaml
+++ b/config/jobs/update-dbg/update-dbg.yaml
@@ -1,0 +1,41 @@
+periodics:
+  - name: update-dbg
+    cron: "0 10 * * 6" # Run every Sunday at 10:00
+    decorate: true
+    extra_refs:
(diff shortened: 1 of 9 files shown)


maxgio92:
As an example, see #685

FedeDP:
I think this *should* be safe to merge, given that old configs are untouched, only new ones are pushed, right?
Basically, only side effect could be that newly added configs are all broken (ie: driverkit is not able to build them).

I see that #685 had some lines removed though: `+106,689 −4 `; i am not able to understand where they were removed though, github is failing bad with so many touched files :)

maxgio92:
Please consider @FedeDP that the PR #685 goal was to test the PR opening feature, from the related test branch with results already in place.

Anyway, this PR would be append-only on distribution (published pre-built drivers). Instead, some configuration could be changed. The latter can't be assumed to be false (e.g. consider a wrong-written configuration that now will be generated).

FedeDP:
Once this is in, i will create a wip PR on top of this to support arm64 drivers build; it should be fairly simple to achieve.

maxgio92:
Do we want @FedeDP to replace `ubuntu-gke` to be treated by Driverkit as a `ubuntu-generic` Ubuntu flavour?
I think we can start simple and then optimize to increase coverage, but I don't have strong opinions on that.

FedeDP:
I think it would be better to fix it now, as it is needed to correctly build that Ubuntu flavour!

maxgio92:
As a side note: I added a `validate` Make target in the DBG, so that we separately validate configurations. In turn the build targets call it.

maxgio92:
@FedeDP @zuc #695 if approved, should be merged before to fix the CI that now is not working

maxgio92:
I agree @FedeDP 

/hold

zuc (approved):

PR #688, opened 2022-04-28
Fix config file path for branchprotector

zuc:

/cc @leogr 
/cc @maxgio92


maxgio92 (approved):
/approve 
thanks @zuc Good catch!

PR #727, opened 2022-06-13
feat(config/clusters): add update-kernels oci image repo

maxgio92:
Signed-off-by: Massimiliano Giovagnoli <me@maxgio.it>


leogr (approved):
good catch

maxgio92:
> good catch

@FedeDP

PR #728, opened 2022-06-13
fix(scripts): use single quotes to escape newline chars in "body" variable

alacuku:
```
Signed-off-by: Aldo Lacuku <aldo@lacuku.eu>
```


zuc (approved):

zuc:
@alacuku nice catch! thanks

PR #730, opened 2022-06-14
fix(images): fix update-kernels image entrypoint python commands.

FedeDP:


maxgio92 (approved):
/approve

PR #736, opened 2022-06-20
cleanup(driverkit/configs): dropped old driver versions.

FedeDP:
As outlined by our `Falco Artifacts Cleanup` proposal, we only need to keep latest 2 versions for Falco stable releases; here, we keep 3 of them: 0.32.0, 0.31.1, 0.31.0.

See: https://github.com/falcosecurity/falco/blob/master/proposals/20200901-artifacts-cleanup.md#drivers


maxgio92 (approved):
/approve

PR #745, opened 2022-06-28
fix(driverkit): do not include generate script in build script to avoid all of its logic

FedeDP:
Instead, just re-define arch_to_driverkit_arch().
It will be removed once we drop support for old tree anyway.


maxgio92 (approved):
/approve

PR #750, opened 2022-07-04
cleanup(config,images): dropped build-libs images

FedeDP:
Because of https://github.com/falcosecurity/libs/pull/453.


FedeDP:
The relevant PR is now merged in libs (together with a fix :D ): https://github.com/falcosecurity/libs/actions/runs/2664065155.


Note: this will need to be force-pushed because master test-infra is watching on images/build-libs to trigger the build-images-build-libs workflow, but the `build-libs` image is no more there in this PR :D

leogr:
> /override build-images-build-libs

The job is not required, so the PR can be merged even without it.

leogr (dismissed):
/approve

leogr (commented):
Hey @FedeDP ,

can we try to split this PR in two?

1. remove the configs but not the `images/build-libs/*` files
2. then, remove `images/build-libs/`

That should do the trick to pass the failing check :point_down: 
https://prow.falco.org/view/s3/falco-prow-logs/pr-logs/pull/falcosecurity_test-infra/750/build-images-build-libs/1546779844036005888

If that doesn't work anyway, we could manually merge it (cc @maxgio92 )

FedeDP:
I will @leogr !

FedeDP:
Done!

maxgio92 (approved):
/approve

PR #757, opened 2022-07-07
chore(configs): updated run-if-changed for dbg jobs splitted by major kernel version.

FedeDP:
Now, each kernelversion job only triggers for its kernelversion configs.


maxgio92 (approved):
/approve

PR #758, opened 2022-07-07
fix(driverkit/utils): fixed scrape_and_generate and generate scripts, ensuring proper kernelversion is actually set

FedeDP:


driverkit/utils/generate
```diff
@@ -47,7 +47,7 @@ if [ -z ${TARGET_VERSION} ]; then
 fi
 
 ensure_kernelversion() {
-    TARGET_KERNEL_VERSION="$(expr match "${TARGET_KERNEL}" '^.*_\(.*\)$')"
+    TARGET_KERNEL_VERSION="$(expr match "${TARGET_KERNEL}" '^.*_\(\d*\)$')"
```
FedeDP, line 50: This was broken for TARGET_KERNEL that contained a `x86_64` (`_64`)

driverkit/utils/scrape_and_generate
```diff
@@ -31,7 +31,7 @@ function generate_from_kernel_releases() {
 			test $DRY_RUN == "true" || \
 				make -C "$(dirname $0)/.." \
 				-e TARGET_ARCH="${ARCH}" \
-				-e TARGET_KERNEL="${release}" \
+				-e TARGET_KERNEL="${release}_${version}" \
```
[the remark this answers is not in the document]
maxgio92: Good catch! Thanks @FedeDP

driverkit/utils/generate
maxgio92: Good catch also here @FedeDP! Thanks.
It would be needed to test also if a valid version is kept, and not always being defaulted to "1".

FedeDP: You were right; it wasn't working and always default to "1".
Now i take for granted that a kernelversion is always present in the crawler output (that holds true of course!).
This way we can actually simplify the code a bit.
Moreover, there is no simple way to distinguish between a "_%kv" and "x86_64" (when kv is missing), therefore the check would have need to be even more convoluted.

FedeDP: If any kernelversion is not present, we can take it as a bug on the crawler.

driverkit/utils/generate
```diff
@@ -52,6 +62,49 @@ compute_kernelrelease() {
     TARGET_KERNEL_RELEASE="${TARGET_KERNEL%_${TARGET_KERNEL_VERSION}}"
 }
 
+check_outputs() {
+    regex='([0-9]+)\.([0-9]+)'
+    if [[ "${TARGET_KERNEL}" =~ ${regex} ]]
+    then
+        maj=${BASH_REMATCH[1]}
```
FedeDP, line 69: Take maj and minor kernel version from the target kernel.

driverkit/utils/generate
```diff
@@ -52,6 +62,49 @@ compute_kernelrelease() {
...
+        min=${BASH_REMATCH[2]}
+
+        # Use correct array for target arch
+        declare -n arrayname=${TARGET_ARCH}_MINIMUM_KVER
```
FedeDP, line 73: Use `_MINIMUM_KVER` array for TARGET_ARCH.

driverkit/utils/generate
```diff
@@ -52,6 +62,49 @@ compute_kernelrelease() {
...
+        ctr=0
+        for min_kver in "${arrayname[@]}" ; do
+            KEY="${min_kver%%:*}"
+            VALUE="${min_kver##*:}"
+
+            if [[ "${VALUE}" =~ ${regex} ]]
+            then
+                req_maj=${BASH_REMATCH[1]}
+                req_min=${BASH_REMATCH[2]}
```
FedeDP, line 82: Take required major and minor versions, ie: the minimum kernel version where we support eBPF and kmod.

driverkit/utils/generate
```diff
@@ -115,13 +174,20 @@ generate_yamls() {
...
+    fi
     if [[ -n "${TARGET_HEADERS}" ]]; then
         echo "kernelurls: ${TARGET_HEADERS}" >> ${FILE}
-    fi   
+    fi
 }
 
 compute_kernelversion
 compute_kernelrelease
+check_outputs || exit 0
```
FedeDP, line 192: As said above: do not propagate errors up to the caller.

driverkit/utils/generate
```diff
@@ -46,16 +44,11 @@ if [ -z ${TARGET_VERSION} ]; then
...
 
-ensure_kernelversion() {
-    TARGET_KERNEL_VERSION="$(expr match "${TARGET_KERNEL}" '^.*_\(\d*\)$')"
-
-    # If kernel version is missing
-    test -z $TARGET_KERNEL_VERSION &&\
-	    TARGET_KERNEL_VERSION=$DEFAULT_KERNEL_VERSION &&\
-	    TARGET_KERNEL="${TARGET_KERNEL}_${TARGET_KERNEL_VERSION}"
+compute_kernelversion() {
+    TARGET_KERNEL_VERSION=$(echo "${TARGET_KERNEL##*_}")
```
FedeDP, line 58: Assume `TARGET_KERNEL` has always the `_$kernelversion` suffix.
This holds true for kernel-crawler; and if it weren't, then it would be a bug on the crawler side.

FedeDP:
Here is a diff of files between this branch and master; you will notice:
* Lots of lines like:
```
D	driverkit/config/2.0.0+driver/aarch64/centos_4.18.0-147.8.1.el8_1.aarch64.yaml
A	driverkit/config/2.0.0+driver/aarch64/centos_4.18.0-147.8.1.el8_1.aarch64_1.yaml
...
D	driverkit/config/2.0.0+driver/x86_64/ubuntu-generic_4.15.0-111_1.yaml
A	driverkit/config/2.0.0+driver/x86_64/ubuntu-generic_4.15.0-111_112.yaml
```
That means the fix in the first 2 commits is actually working.
* lots of configs modified -> probe output removed (third commit)
* Some configs removed without none added -> because they were already present in test-infra before #739 was merged, but then #739 wrongly created same configs but without the "_$kernelrelease" suffix, doubling them.
Indeed, we have 4919 x86_64 configs now, vs 5447 on master.

[diff.txt](https://github.com/falcosecurity/test-infra/files/9071845/diff.txt)

FedeDP:
> /test build-new-drivers-centos-2-presubmit
/test build-new-drivers-centos-4-presubmit
/test build-new-drivers-centos-5-presubmit

I think it is better to just force push :/ testing the build for centos takes ages, literally 20hrs or so (for each centos).
We will build centos in any case, on master.

leogr (approved):

driverkit/utils/generate
maxgio92: As discussed, I agree. I resolve this conversation :-)

PR #759, opened 2022-07-11
chore(config): run update-dbg job as soon as new kernel lists are merged in main in kernel-crawler repo

FedeDP:
This allows us to better manage the cross dependency between the update-dbg and update-kernels jobs:
update-kernels will open the PR on kernel-crawler, and as soon as someone approves it (and then it gets merged),
the update-dbg job will trigger.

In the end, we are now able to eg:
* avoid triggering the update-dbg jobs should we don't want to, by just avoiding merging the update-kernels created PR on kernel-crawler repo
* trigger the update-dbg job as quick as possible by merging the kernel-crawler auto PR every week (thus triggering update-dbg weekly).

### DIFF
--- a/config/jobs/update-dbg/update-dbg.yaml
+++ b/config/jobs/update-dbg/update-dbg.yaml
@@ -1,41 +1,44 @@
-periodics:
-  - name: update-dbg
-    cron: "0 10 5,20 * *"
-    decorate: true
-    extra_refs:
-    # Check out the falcosecurity/test-infra repo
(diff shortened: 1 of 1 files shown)


maxgio92:
Good idea @FedeDP. Thanks!

maxgio92 (approved):
/approve

FedeDP:
/hold

Let's wait for the 0.32.1 drivers first :)

PR #766, opened 2022-07-15
fix(azure-distro): Fix wildcard to be able to use with azure ubuntu

Lowaiz:

Removing the '-' of the wildcards search to be able to access every config.

And I have a question about the azure's ubuntu config: All configs I've seen have a target set to `ubuntu-azure` but the `driverkit`  tools doesn't support it. So it crash when I use them. Is this the expected setup ?


maxgio92:
Hi @Lowaiz thank you!
/lgtm

/assign @FedeDP 

About Driverkit, unfortunately the logic is not implemented yet. There is a work in progress [here](https://github.com/falcosecurity/driverkit/issues/144).

FedeDP:
I don't quite like it, because filtering for eg:
TARGET_KERNEL=4.10.9 would also take 4.10.{90,91...900,901} and that is an issue IMHO.
The best approach is https://github.com/falcosecurity/driverkit/pull/161 and just use "ubuntu" as a target for new Ubuntu related driverkit configs.
We can then use sed to replace any ubuntu-generic/Ubuntu-aws confg with just "ubuntu"  (but it is not mandatory because that PR will maintain backward compatibility!)

Lowaiz:
> TARGET_KERNEL=4.10.9 would also take 4.10.{90,91...900,901} and that is an issue IMHO

I agree. But with the actual wildcard and a `TARGET_KERNEL` set to `5.4.0-1044-azure_46`, i.e., make won't find any config to apply. You can pass a `TARGET_KERNEL` set to `5.4.0-1044` with a `TARGET_DISTRO` set to the ubuntu flavor, but the actual driverkit won't let a target distro set to an ubuntu flavor other then aws, so it's quite annoying to move around to deploy drivers the best way. Is there any chance to add the azure option to the driverkit, at least temporarily, before the full flavor compatibility release ?

Anyway, I'll take a look to the falcosecurity/driverkit#161 today or tomorrow.

Lowaiz:
After some test with the falcosecurity/driverkit#161, I think there will be some little problems with wildcard. As the verifying script works, the config files should be named like that `"${target}_${kernelrelease}_${kernelversion}.yaml"` but with the new way of driverkit behavior, the kernel release can have an extra version (i.e. `5.4.0-1044-azure`) and the current pattern in the Makefile `${TARGET_DISTRO}_${TARGET_KERNEL}-*` doesn't make the difference between the release and the kernel version, and with the dash, if we plug an flavored ubuntu release, the file won't be detected. 
I guess we should make the wildcard the same format as the script, to be able to select with precision the kernel release and the kernel version.
WDYT ?

FedeDP:
@Lowaiz it is quite hard to grasp it like that :) 
Can you make an example?
Thanks!

Lowaiz:
To make it more clear, I want to be able to generate a config file on the fly and use it to build the driver when I want to deploy Falco.
So I tried the `utils/generate` script with this kind of params: 
```
 ./utils/generate -a x86_64 -k 5.4.0-1044-azure_46 -d ubuntu -v 39ae7d40496793cf3d3e7890c9bbdc202263836b
```
Where `5.4.0-1044-azure` is the kernel release and the `46` is the kernel version.
This command generate a file called `ubuntu_5.4.0-1044-azure_46.yaml`.
Then I run the Makefile to start building the driver:
```
make -e TARGET_KERNEL=5.4.0-1044-azure_46 -e TARGET_DISTRO=ubuntu -e TARGET_VERSION=39ae7d40496793cf3d3e7890c9bbdc202263836b specific_target
```
But since the pattern to look at all configs available is `config/${TARGET_VERSION}*/${TARGET_DISTRO}_${TARGET_KERNEL}-*` so with our values `config/39ae7d40496793cf3d3e7890c9bbdc202263836b*/ubuntu_5.4.0-1044-azure_46-*`, the wildcard can't find the file because of the trailing '-' at the end of the pattern.

My suggestion would be to explicitly split the kernel release and version from the input of the Makefile to build a more precise pattern.

FedeDP:
Uh i see! It makes sense! 
I think the hardest part here is not breaking current behavior, but we can try! Are you willing to update this PR?
Btw we need to have https://github.com/falcosecurity/driverkit/pull/161 merged first :)

Lowaiz:
> Are you willing to update this PR?

Of course, I'll take a moment to try before tomorrow.

Lowaiz:
@maxgio92 or @FedeDP

Can you take a look at the backward compatibility part ?
I tried the few example from the README and they works (except there is no config for debian with kernel 5.9) , but I don't know any other edge cases.

Lowaiz:
Oh, I guess there is still  this problem:
> TARGET_KERNEL=4.10.9 would also take 4.10.{90,91...900,901}

The only clean way that come to me to handle this case, is that we remove the `*` and we let the user adding it.

FedeDP:
I need to study this well, sorry if it will take a bit of time; fact is, we also have #772 now, and the risk of breaking anything is skyhigh imho!

Lowaiz:
I just thought about this: 
We could add another input param to the makefile, like `TARGET_KERNEL_RELEASE_WILDCARD`, as a boolean, set by default to `true`, and only if this param is set to `true`, we append the `*` to the release.

Lowaiz:
I really think that the `TARGET_KERNEL_RELEASE_WILDCARD` could be a great option, as it reproduce the targeted behavior (ie, targeting kernel version 5.4.0-xxx) but extend to possibility to filter kernel version and fix the bug of the current wildcard.

Lowaiz:
Change made and after some tests it seems ok

maxgio92:
Hey @Lowaiz thank you.

I made some tests and tried to simplify the DBG logics. 
Basically is that, whether we pass a kernel name or (kernel release and/or kernel version), it will work.

#### Suggested change

My suggestion is to simplify the PR with something like this:
```diff
diff --git a/driverkit/Makefile b/driverkit/Makefile
index b83648feb..46a74ae04 100644
--- a/driverkit/Makefile
+++ b/driverkit/Makefile
@@ -18,7 +18,9 @@ VERSIONS := $(patsubst config/%,%,$(sort $(dir $(wildcard config/*/))))
 VERSIONS := $(VERSIONS:/=)
 TARGET_VERSION ?= *
 TARGET_DISTRO ?= *
-TARGET_KERNEL ?= *
+TARGET_KERNEL_RELEASE ?= *
+TARGET_KERNEL_VERSION ?= *
+TARGET_KERNEL ?= $(TARGET_KERNEL_RELEASE)_$(TARGET_KERNEL_VERSION)
 TARGET_ARCH ?= *
 TARGET_ARCHS := $(if $(call eq,$(TARGET_ARCH),*),$(ALL_ARCHS),$(TARGET_ARCH))
 TARGET_HEADERS ?=
@@ -42,9 +44,9 @@ specific_target: specific_target_old specific_target_new
 # because specific_target_old will always build any old config.
 # Not a big deal because we were not going to filter for TARGET_ARCH in any case.
 # TODO: rm once old format support is dropped
-specific_target_old: $(patsubst config_%,%,$(subst /,_,$(wildcard config/${TARGET_VERSION}*/${TARGET_DISTRO}_${TARGET_KERNEL}-*)))
+specific_target_old: $(patsubst config_%,%,$(subst /,_,$(wildcard config/${TARGET_VERSION}*/${TARGET_DISTRO}_${TARGET_KERNEL}.yaml)))
 # new fmt has $arch subfolders
-specific_target_new: $(patsubst config_%,%,$(subst /,_,$(wildcard config/${TARGET_VERSION}*/${TARGET_ARCH}/${TARGET_DISTRO}_${TARGET_KERNEL}-*)))
+specific_target_new: $(patsubst config_%,%,$(subst /,_,$(wildcard config/${TARGET_VERSION}*/${TARGET_ARCH}/${TARGET_DISTRO}_${TARGET_KERNEL}.yaml)))
 
 prepare: $(addprefix prepare_,$(VERSIONS))
 publish: publish_s3 # alias publish_s3; artifactory is no more supported
```

#### Tests

I tested your use cases with this patch with success, and you can reproduce them with:

```bash
#!/usr/bin/env bash

set -x

# -- New Driverkit with general Ubuntu support

# Ubuntu Azure
./utils/generate -a x86_64 -k 5.4.0-1044-azure_46 -d ubuntu -v 39ae7d40496793cf3d3e7890c9bbdc202263836b
make \
-e TARGET_KERNEL_RELEASE=5.4.0-1044-azure \
-e TARGET_KERNEL_VERSION=46 \
-e TARGET_DISTRO=ubuntu \
-e TARGET_VERSION=39ae7d40496793cf3d3e7890c9bbdc202263836b \
specific_target

make \
-e TARGET_KERNEL=5.4.0-1044-azure_46 \
-e TARGET_DISTRO=ubuntu \
-e TARGET_VERSION=39ae7d40496793cf3d3e7890c9bbdc202263836b \
specific_target

# -- Legacy Driverkit with specific Ubuntu support

# Ubuntu Azure
./utils/generate -a x86_64 -k 5.4.0-1044-azure_46 -d ubuntu-azure -v 39ae7d40496793cf3d3e7890c9bbdc202263836b
make \
-e TARGET_KERNEL_RELEASE=5.4.0-1044-azure \
-e TARGET_KERNEL_VERSION=46 \
-e TARGET_DISTRO=ubuntu-azure \
-e TARGET_VERSION=39ae7d40496793cf3d3e7890c9bbdc202263836b \
specific_target

make \
-e TARGET_KERNEL=5.4.0-1044-azure_46 \
-e TARGET_DISTRO=ubuntu-azure \
-e TARGET_VERSION=39ae7d40496793cf3d3e7890c9bbdc202263836b \
specific_target
```

What do you think?

Lowaiz:
Hey @maxgio92 

This part is clearly better than mine:
```
TARGET_KERNEL_RELEASE ?= *
TARGET_KERNEL_VERSION ?= *
TARGET_KERNEL ?= $(TARGET_KERNEL_RELEASE)_$(TARGET_KERNEL_VERSION)
```

But for the wildcard, we are loosing the `-*`, that was made to put any semver kernel version (i.e. 5.4.0) and build every kernel release and version of this semver. (i.e. 5.4.0-122_138)

FedeDP:
> But for the wildcard, we are loosing the -*, that was made to put any semver kernel version (i.e. 5.4.0) and build every kernel release and version of this semver. (i.e. 5.4.0-122_138)

But i think one can achieve the same by passing:
```
make \
-e TARGET_KERNEL_RELEASE=5.4.0-* \
-e TARGET_KERNEL_VERSION=138 \
-e TARGET_DISTRO=ubuntu \
-e TARGET_VERSION=39ae7d40496793cf3d3e7890c9bbdc202263836b \
specific_target
```

Lowaiz:
> But i think one can achieve the same by passing:

I fully agree, that even was my argument at the middle of this discussion.
 
> The only clean way that come to me to handle this case, is to remove the * after TARGET_KERNEL_RELEASE and we let the user adding it.
An argument for that is that even the README give us an example of a target with a *, so the presence of * in the pattern is kinda redundant.

But this is causing breaking change, that you wanted to avoid.

FedeDP:
I see, i was possibly wrong all the time then :) 
The only usage we have is just filtering for major kernel version, like
```
make \
-e TARGET_KERNEL=5.*  \
-e TARGET_DISTRO=ubuntu \
-e TARGET_VERSION=39ae7d40496793cf3d3e7890c9bbdc202263836b \
specific_target
```

driverkit/Makefile
```diff
@@ -18,7 +18,11 @@ VERSIONS := $(patsubst config/%,%,$(sort $(dir $(wildcard config/*/))))
 VERSIONS := $(VERSIONS:/=)
 TARGET_VERSION ?= *
 TARGET_DISTRO ?= *
-TARGET_KERNEL ?= *
+TARGET_KERNEL_RELEASE ?= *
+TARGET_KERNEL_VERSION ?= *
+TARGET_KERNEL ?= ${TARGET_KERNEL_RELEASE}_${TARGET_KERNEL_VERSION}
+TARGET_KERNEL_RELEASE_WILDCARD ?= true
```
FedeDP, line 24: Leftover? :)

Lowaiz: oopsy

FedeDP (approved):
IMHO this LGTM! 
I cannot approve, but my 2c anyway :)

maxgio92 (approved):
/approve

Thank you so much @Lowaiz! 🤗

PR #772, opened 2022-07-20
Update for Driverkit's new Ubuntu Support

EXONER4TED:
Signed-off-by: Logan Bond <lbond@secureworks.com>

I ran a sed against the whole of ubuntu Driverkit configs to change the target to `target: ubuntu`:
```bash
#!/bin/bash

for file in $(find ./config/ -name *ubuntu*)
do
        cat $file
        gsed -i 's/target: ubuntu.*/target: ubuntu/g' $file
        cat $file

done
```

I then validated the change with the following:
```
╰─❯ find . -name \*ubuntu\* -exec cat {} + | grep target | sort -n | uniq -c
  10621 target: ubuntu
```

Aside from changing 10k files, that's all I did 😅 

This PR updates the driverkit configs for Ubuntu to use the new `target: ubuntu` format. It leaves everything else the same.


FedeDP:
Hi Logan! Thanks for this PR 🙂
I think this should be on hold until a new driverkit version is released, to avoid breaking anything!

leogr:
Hey @EXONER4TED 

Thank you for catching this! :smile: 

It SGTM.

AFAIK, before merging this we have to wait for a new release of driverkit (@FedeDP right?).

So, on hold for now.

/hold

FedeDP:
I also think that config files should be renamed accordingly: as of now, they are still named "ubuntu-$flavor_$kernelversion"; they should be named "ubuntu_$kernelversion" (these are the new names that would be generated by `generate` utils https://github.com/falcosecurity/test-infra/blob/master/driverkit/utils/generate). BUT we also need the config names to keep the flavor because we have jobs that only run for ubuntu-generic and ubuntu-aws, as you can see: https://github.com/falcosecurity/test-infra/tree/master/config/jobs/build-drivers .

I am not sure how to proceed here: either we use ubuntu_$kernelversion for all (therefore transparently support any ubuntu flavor), or we update `generate` script to manage flavors for ubuntu only, to output "ubuntu-$flavor_$kernelversion" config files. In the latter case, we would have to manually add any new ubuntu flavor `build-drivers` job config (https://github.com/falcosecurity/test-infra/tree/master/config/jobs/build-drivers).
In the former case, instead, the issue is that the build would take ages possibly longer that 1day that is our current test-infra timeout for jobs: https://github.com/falcosecurity/test-infra/blob/master/config/config.yaml#L34.

EXONER4TED:
Ah, sorry, I wasn't quite sure if it was as simple as I'd hoped it would be. I'm not super familiar with how this build grid works.

If it's of any help to you all, in the naming scheme, here is what we use for the modules we build internally:
```
    filename = "scwx_#{config['target']}_#{config['kernelrelease']}_#{config['kernelversion']}.ko"
```
Since the kernel release should contain the flavor on the end of the string, we found it redundant and made the filenames longer than they needed to be. And since the new target update is just `ubuntu`, then `config['target']` should just be `ubuntu` each time (we are actually capturing the `ubuntu-generic` and `ubuntu-aws` targets today and chopping off the flavors to rename them to just `ubuntu`).

That to say...our first build jobs did take a very long while... but the good news is that was a one-time load as subsequent runs don't rebuild anything already existing.

Let me know what you'd like me to do though, and I'll be happy to make changes to scripts/logic wherever (or do nothing, if you'd prefer) 😄

maxgio92:
Sorry, I missed this PR https://github.com/falcosecurity/driverkit/pull/161. Got it.
If the `kernelrelease` is always providing the `flavor`, then consumers (i.e. Driverkit) can take decisions.

I see the issue of dropping off the flavor from configuration filenames and I'd take it for a DBG CI organization.
But I agree with @EXONER4TED as the flavor is redundant, if it always comes with the `kernelrelease` as a suffix. I just would like to be sure this rule is true, otherwise it would be an assumption we make. Am I'm missing something?

EXONER4TED:
You're not missing something 😄 I used kernel-crawler as input for my change to Driverkit. Every instance of an Ubuntu discovered kernel release came with the `flavor` attached to the end (which, in my experience, matches directly out of `uname -r` on Ubuntu systems).

The one catch was `generic` kernels, which just don't have a flavor in kernel-crawler's output. To me, that's actually odd, as running `uname` on an Ubuntu "generic" machine actually does show `-generic` on the end of the kernel release.

Regardless, to get around that, I just defaulted the flavor to `generic` if unable to parse a flavor off the end of the kernel release. I also updated kernel-crawler to default to `generic` as well for kernel releases where it also does not have a flavor.

To me, the big assumption here is that Ubuntu doesn't change their naming scheme, but I would think that would mean bigger changes for kernel-crawler/driverkit anyway...

FedeDP:
> To me, the big assumption here is that Ubuntu doesn't change their naming scheme, but I would think that would mean bigger changes for kernel-crawler/driverkit anyway...

I agree with this! Again, my biggest concern is how to properly manage load of our cluster; at the moment we are filtering jobs by ubuntu flavor; but we would loose the ability to do so; i think that filtering for kernel major version only is going to be very heavy on our resources.
But we **need** this IMHO.

FedeDP:
I would like to also better understand how this one and #766  will work together; they are somewhat touching the same areas and i fear we would break something :/

Lowaiz:
> how this one and https://github.com/falcosecurity/test-infra/pull/766 will work together

The current state of #766 permit the split of kernel version and release and it's first goal was to fix a bug where ubuntu-azure config wasn't found by this wildcard when you give a precise kernel release and version, and this one (and the driverkit update) permits to not specify the flavor of ubuntu into the target distribution, and keep it in the kernel release.

So together, they simplify the way of targeting a config, by giving the possibilty to give only `ubuntu` as targeted distro and a full kernel release to the Makefile to build the drivers. 

> i fear we would break something 

Apart the wildcard problem mentioned in #766, I can't think of breaks.

FedeDP:
On a second though, the flavor will always be present, right?
But its string will be at the end of the kernelrelease, like
`ubuntu-5.10.0-aws_49.yaml`

Am i correct? With this, we can still keep current flavor-based load management (even if this practically denies us from supporting any ubuntu flavor, though).

Lowaiz:
> Am i correct? 

Yes

> we can still keep current flavor-based load management 

What do you mean by that ?

FedeDP:
> What do you mean by that ?

I mean, today we have [ubuntu-aws](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-drivers/build-new-ubuntu-aws.yaml#L9) and [ubuntu-generic](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-drivers/build-new-ubuntu-generic.yaml#L9) specific jobs; we could retain same behavior by updating the regex, because the flavor is still present, but it it no more part of the `target`; instead, it is part of the kernelrelease.

Lowaiz:
> we have [ubuntu-aws](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-drivers/build-new-ubuntu-aws.yaml#L9) and [ubuntu-generic](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-drivers/build-new-ubuntu-generic.yaml#L9) specific jobs

Oh I never used these config. With a quick look at them, I guess you are correct, buy why did you split that into 2 separated  jobs while they are exactly the same, except the 'aws' occurrence are replaced by 'generic' ?

FedeDP:
Load management of our cluster :) It is better to have like: N small jobs instead of a single giant one.
If you look carefully, we are also splitting ubuntu configs by kernel major version in those files, for the exact same reason!

Lowaiz:
After taking a look to the behavior of these jobs, I understand a bit more you fear of breaking something.
With the #766 I guess we'll need to modify this [script](https://github.com/falcosecurity/test-infra/blob/master/images/build-drivers/build-drivers.sh)

FedeDP:
There is some more concern about this PR: basically, [falco-driver-loader](https://github.com/falcosecurity/falco/blob/master/scripts/falco-driver-loader#L131) embeds logic to compute where to lookup for the driver on download.falco.org. As you can see, `ubuntu-generic` and `ubuntu-aws` are embedded there.

But i still think this is the correct way forward, therefore i propose this plan:
* in this PR, update also the `output:` paths for each ubuntu config, that right now are wrong (ie: they are still using ubuntu-generic and ubuntu-aws as targets for output kmod/probe paths)
* open a PR on falco to fix the falco-driver-loader logic (basically `TARGET_ID="${OS_ID}"` is enough for ubuntu case)
* Finally, merge the [build-new-ubuntu-aws](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-drivers/build-new-ubuntu-aws.yaml) and [build-new-ubuntu-generic](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-drivers/build-new-ubuntu-generic.yaml] in a single `build-new-ubuntu` job.

@EXONER4TED are you willing to take on this duties? :) 
They're just small PRs, but i know you were expecting a lower effort :/ sorry for that!

maxgio92:
Agree on all @FedeDP.
> Finally, merge the [build-new-ubuntu-aws](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-drivers/build-new-ubuntu-aws.yaml) and [build-new-ubuntu-generic](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-drivers/build-new-ubuntu-generic.yaml) in a single build-new-ubuntu job.

I'd add also [rebuild-ubuntu-aws](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-drivers/rebuild-all-ubuntu-aws.yaml) [rebuild-ubuntu-generic](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-drivers/rebuild-all-ubuntu-generic.yaml) and jobs.

Again, sorry @EXONER4TED for this and thank you for all this lot of work you're doing!

FedeDP:
> Finally, merge the [build-new-ubuntu-aws](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-drivers/build-new-ubuntu-aws.yaml) and [build-new-ubuntu-generic](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-drivers/build-new-ubuntu-generic.yaml) in a single build-new-ubuntu job.

On a second thought, after a discussion with @leogr , we decided to avoid merging the configs because test-infra would not been able to build all ubuntu flavors in a single job; instead, we will be creating new configs for each ubuntu flavor we want to support.

Moreover, i am not sure about first point too (ie: updating `output:` in this PR) because we don't want to rebuild all drivers now (changing the output name will trigger a full rebuild);
all in all, i think this PR is quite good as is, we will need to add a workaround in `scrape_and_generate` test-infra driverkit script that:
* for currently supported driverversions, updates `output` values by adding the `-flavor` part, so that newly crawled kernels will still work for currently supported versions (and falco-driver-loaders)
* for next driverversion, the workaround won't be needed; 
* then, after 3 versions (ie: when we only support new `ubuntu` target and output paths) we can drop the workaround

I'll have to think about this a little bit more, but it makes sense for now. 
I will be back asap with further thoughts :)

FedeDP:
In the end, we decided that the best thing we could do is just:
* using proper ubuntu-$flavor as target in kernel-crawler and driverkit
* allow driverkit to manage any ubuntu-$flavor (as just "ubuntu" target internally)
* update falco-driver-loader to avoid fallbacking at "ubuntu-generic" when not in "ubuntu-aws" env; instead, use the correct ubuntu flavor

See:
* https://github.com/falcosecurity/falco/pull/2178
* https://github.com/falcosecurity/kernel-crawler/pull/38
* https://github.com/falcosecurity/driverkit/pull/201

We will need new prow jobs for any ubuntu flavor we want to support!

FedeDP:
If @EXONER4TED is ok, i think this can be closed.

EXONER4TED:
Hey! Apologies for the exceptional absence on my part - work life and personal life got a bit crazy busy there for a while 😅 

I am up for whatever work you guys need me to do. I'll close this PR with this comment, but let me know if there is anything more I can do to help. This one was a bit of a hard one because I am not familiar with the build systems and automation you guys have under the hood here haha...

I will work to be less absent. Life just got a little crazy there.

FedeDP:
No problem @EXONER4TED , you already did a great job :) 
Thanks!

PR #791, opened 2022-08-01
fix(config/plugins.yaml): enable maintainers team for milestone plugin

jasondellaluce:
Signed-off-by: Jason Dellaluce <jasondellaluce@gmail.com>

This fixes the org teams mapping for the milestone plugin, which have not been updated after the recent org.yaml updates, and prevented the `/milestone` command to be used in all repos for non-admin maintainers.


leogr (approved):
Good catch!

/approve

PR #795, opened 2022-08-04
fix(utils-generate): Bringing back the compute functions

Lowaiz:
generate kernel from release and version, generate release and version from kernel.

A previous update messed up a bit with params, getting exit due to at least one of `TARGET_KERNEL`/`TARGET_KERNEL_RELEASE`/`TARGET_KERNEL_VERSION` that was not set.
Bringing back some old functions to make it more breaking proof.

If `TARGET_KERNEL` is set, but not the release or version, we extract them from `TARGET_KERNEL`.
if `TARGET_KERNEL` is not set, we compute it with the release and version.

### DIFF
--- a/driverkit/utils/generate
+++ b/driverkit/utils/generate
@@ -70,6 +70,7 @@ compute_kernelrelease() {
 # If target kernel is not empty, compute release and version
 # Else, compute it with release and version
 if [ -n "${TARGET_KERNEL}" ]; then
+  echo "The '-k TARGET_KERNEL' option is deprecated, prefer the '-r TARGET_KERNEL_RELEASE -t TARGET_KERNEL_VERSION' combo"
   if [ -z "${TARGET_KERNEL_VERSION}" ]; then
     compute_kernelversion
   fi
@@ -78,7 +79,7 @@ if [ -n "${TARGET_KERNEL}" ]; then
   fi
 else
   if [ -z "${TARGET_KERNEL_RELEASE}" ] || [ -z "${TARGET_KERNEL_VERSION}" ]; then
-    echo "TARGET_KERNEL, TARGET_KERNEL_RELEASE, TARGET_KERNEL_RELEASE can't be empty at the same time"
+    echo "Options '-k TARGET_KERNEL', '- r TARGET_KERNEL_RELEASE', '-t TARGET_KERNEL_RELEASE' can't be empty at the same time"
     exit 1
   fi
   compute_kernel
--- a/driverkit/utils/generate
+++ b/driverkit/utils/generate
@@ -70,11 +70,11 @@ compute_kernelrelease() {
 # If target kernel is not empty, compute release and version
 # Else, compute it with release and version
 if [ -n "${TARGET_KERNEL}" ]; then
-  if [ -z "${TARGET_KERNEL_RELEASE}" ]; then
-    compute_kernelrelease
-  fi
   if [ -z "${TARGET_KERNEL_VERSION}" ]; then
     compute_kernelversion
+  fi
+  if [ -z "${TARGET_KERNEL_RELEASE}" ]; then
+    compute_kernelrelease
   fi
 else
   if [ -z "${TARGET_KERNEL_RELEASE}" ] || [ -z "${TARGET_KERNEL_VERSION}" ]; then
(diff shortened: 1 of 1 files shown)


driverkit/utils/generate
```diff
@@ -45,29 +45,45 @@ while getopts ":a:k:d:v:h:c:r:t:" arg; do
...
+}
+
+compute_kernelrelease() {
+    TARGET_KERNEL_RELEASE="${TARGET_KERNEL%_${TARGET_KERNEL_VERSION}}"
+}
+
+# If target kernel is not empty, compute release and version
+# Else, compute it with release and version
+if [ -n "${TARGET_KERNEL}" ]; then
+  if [ -z "${TARGET_KERNEL_RELEASE}" ]; then
```
maxgio92, line 73: I think @Lowaiz `TARGET_KERNEL_VERSION` needs to be compiled first

driverkit/utils/generate
```diff
@@ -45,29 +45,45 @@ while getopts ":a:k:d:v:h:c:r:t:" arg; do
...
-fi
-if [ -z ${TARGET_VERSION} ]; then
-    echo "TARGET_VERSION can't be empty"
+
+compute_kernel() {
+    TARGET_KERNEL="${TARGET_KERNEL_RELEASE}_${TARGET_KERNEL_VERSION}"
+}
+
+compute_kernelversion() {
+    TARGET_KERNEL_VERSION=$(echo "${TARGET_KERNEL##*_}")
```
alacuku, line 63: This one will not work in some cases. In the minikube case we have: 

- kernelversion: "1_1.26.0"
- kernelrelease: "5.10.57"
The TARGET_KERNEL will be "5.10.57_1_1.26.0". The compute_kernelversion and compute_kernelrelease will return wrong values.

Can I ask why do you need to get the version and release from the TARGET_KERNEL when you can just pass them separately to the `generate` script?

Lowaiz: > Can I ask why do you need to get the version and release from the TARGET_KERNEL when you can just pass them separately to the generate script?

 Not my direct need, it's to avoid breaking change on the behavior of the script. We can mark the TARGET_KERNEL as deprecated. 
 
> This one will not work in some cases. In the minikube case we have:
>
>    kernelversion: "1_1.26.0"
>    kernelrelease: "5.10.57"
>
> The TARGET_KERNEL will be "5.10.57_1_1.26.0". The compute_kernelversion and compute_kernelrelease will return wrong values.

If TARGET_KERNEL is computed by the release and the version, the compute_kernelversion and compute_kernelrelease won't be call, as you already gave to the script the information.

driverkit/utils/generate
Lowaiz: Nice catch

driverkit/utils/generate
alacuku: As far as I know, the TARGET_KERNEL is always expected to be computed by using the release and the version. The `falco-driver-loader` relies on this logic when computing the falco driver URL: https://github.com/falcosecurity/falco/blob/c81f3fc87ed622d2f35f744a19263a7d1f262e2c/scripts/falco-driver-loader#L237-L238
I think that we have other scripts that rely on the same logic.

As I said before the old logic does not work when the TARGET_KERNEL has more than one `_` character. If we really need to preserve the old behavior then we need to find a clever way to compute the `version` and `release` from the TARGET_KERNEL.

Anyway, this is just my opinion.

Lowaiz: I agree about the fact the a better way could be found and that your case doesn't work with only TARGET_KERNEL set, but the point of this MR is to bring back the possibilty to give only the TARGET_KERNEL, like before, computing the release and the version, even if it breaks on your case, as it's a new feature with this split of the params, we can keep TARGET_KERNEL like this, as deprecated.

I prefer the new way with release and version as params, but it's not the point :D

If you use `TARGET_KERNELRELEASE` / `TARGET_KERNELVERSION`, you can just ignore `TARGET_KERNEL`.

maxgio92: I agree with you @alacuku, but unfortunately I'm not sure there is a deterministic way to calculate the version from the full kernel name.

Anyway, as @Lowaiz said, this is just to be backward-compatible, as this was the old way to generate the configuration (without explicit release + version input params).

I propose then to move to deprecate TARGET_KERNEL so that we can remove the non-deterministic computations, like those.

maxgio92 (approved):
/approve

PR #796, opened 2022-08-04
feat(driverkit-kube): Build utils can now run on kube by DRIVERKIT_BACKEND variable

Lowaiz:

Everything in the title


maxgio92:
Thank you @Lowaiz.
This could enable a refactoring of the current infra architecture, and be discussed.

In this scenario, we need to understand how those pods would be recognized by Prow related to `ProwJob`s.

PR #797, opened 2022-08-05
feat(driverkit-kube): Build utils can now run on kube

Lowaiz:

you merged the previous a bit too fast :D

### DIFF
--- a/driverkit/utils/build
+++ b/driverkit/utils/build
@@ -8,6 +8,7 @@ source "${currentdir}/parseyaml"    # create_variables
 input="$1"
 IFS=/ read -a input_parts <<< ${input}
 driver_version="${input_parts[1]}"
+DRIVERKIT_BACKEND="${DRIVERKIT_BACKEND:-docker}"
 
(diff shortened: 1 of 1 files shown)


driverkit/utils/build
```diff
@@ -72,5 +72,8 @@ fi
 if [[ ${should_build} == "true" ]]; then
     mkdir -p output/${driver_version}/${arch}
     # if we have an architecture in our config path, force use it
+    if [[ -z ${DRIVERKIT_BACKEND} ]]; then
```
maxgio92, line 75: Hi @Lowaiz, good catch! Thank you.

I'd rather use [bash expansion](https://www.gnu.org/software/bash/manual/html_node/Shell-Parameter-Expansion.html) with default value, something like:
```bash
DRIVERKIT_BACKEND= ${DRIVERKIT_BACKEND:-docker}
```
from line 10

maxgio92 (changes requested):

maxgio92 (approved):
/approve

Thanks @Lowaiz!

PR #804, opened 2022-08-12
Add select GCP ubuntu-generic kernels that kernel-crawler does not find

0xIAN:
Signed-off-by: Ian Robertson <ian.robertson@wpengine.com>

This adds kernel support for 94 GCP Compute Engine kernels of the "ubuntu-generic" flavor. These kernels are not being discovered and built properly as part of the automation behind kernel-crawler. This does not include every kernel version of this type, only a select set.


leogr:
/ok-to-test

cc @FedeDP @maxgio92

maxgio92:
Thank you @0xIAN

leogr (dismissed):

IanRobertson-wpe:
I'm unsure why there are some remaining errors for the ubuntu-generic...-gcp drivers. I've just completed re-testing every single one of these locally, and they are building kernel and probes correctly with driverkit version v0.9.5-0+f1d7e7f. A "-l debug" on driverkit would be needed for the Prow job in order to diagnose further. I tend to suspect a transient issue at play, perhaps an oomkill or a network issue.

leogr:
> I'm unsure why there are some remaining errors for the ubuntu-generic...-gcp drivers. I've just completed re-testing every single one of these locally, and they are building kernel and probes correctly with driverkit version v0.9.5-0+f1d7e7f. A "-l debug" on driverkit would be needed for the Prow job in order to diagnose further. I tend to suspect a transient issue at play, perhaps an oomkill or a network issue.

```json
{"component":"entrypoint","file":"k8s.io/test-infra/prow/entrypoint/run.go:255","func":"k8s.io/test-infra/prow/entrypoint.gracefullyTerminate","level":"error","msg":"Process did not exit before 10m0s grace period","severity":"error","time":"2022-08-23T14:51:28Z"}
```

:(

/test build-new-drivers-ubuntu-generic-4-presubmit

:crossed_fingers:

leogr:
> @maxgio92 It seems the timeouts are continuing. Do you know what is causing the timeout? Should the timeout be extended?

I think so. cc @zuc

IanRobertson-wpe:
Anything I can do here? It seems like the timeout needs to be extended.

zuc:
Actually I’m more inclined to see if we can either split the job into multiple sub-jobs and/or verify if all the libs driver versions that we are building are still needed.

Rationale for this is that tweaking the timeout param will result in a blocker for us to evaluate other CI alternatives in the future (e.g. Github Actions), as typically this is not something you can do on hosted solutions.

What do other @falcosecurity/test-infra-maintainers think?

FedeDP:
BTW i don't get why is this failing for timeouts, while eg: #824 passed with no issues (aside from some scheduling issues) :/

FedeDP:
Hi! Another note: kernel-crawler is actually finding these kernels, see eg:
```
{
      "kernelversion": "8",
      "kernelrelease": "4.15.0-1008-gcp",
      "target": "ubuntu-gcp",
      "headers": [
        "http://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-gcp/linux-gcp-headers-4.15.0-1008_4.15.0-1008.8_amd64.deb",
        "http://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-gcp/linux-headers-4.15.0-1008-gcp_4.15.0-1008.8_amd64.deb",
        "http://security.ubuntu.com/ubuntu/pool/main/l/linux-gcp/linux-headers-4.15.0-1008-gcp_4.15.0-1008.8_amd64.deb",
        "http://security.ubuntu.com/ubuntu/pool/main/l/linux-gcp/linux-gcp-headers-4.15.0-1008_4.15.0-1008.8_amd64.deb"
      ]
    },
```
But as you can see from the above json element, the target is `ubuntu-gcp` and not `ubuntu-generic`.
That is, IMHO, correct because it is an ubuntu-gcp.
Therefore, configs are created like:
```
cat ubuntu-gcp_4.15.0-1008-gcp_8.yaml
kernelversion: 8
kernelrelease: 4.15.0-1008-gcp
target: ubuntu-gcp
architecture: amd64
output:
module: output/2.0.0+driver/x86_64/falco_ubuntu-gcp_4.15.0-1008-gcp_8.ko
probe: output/2.0.0+driver/x86_64/falco_ubuntu-gcp_4.15.0-1008-gcp_8.o
kernelurls: ["http://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-gcp/linux-gcp-headers-4.15.0-1008_4.15.0-1008.8_amd64.deb","http://mirrors.edge.kernel.org/ubuntu/pool/main/l/linux-gcp/linux-headers-4.15.0-1008-gcp_4.15.0-1008.8_amd64.deb","http://security.ubuntu.com/ubuntu/pool/main/l/linux-gcp/linux-headers-4.15.0-1008-gcp_4.15.0-1008.8_amd64.deb","http://security.ubuntu.com/ubuntu/pool/main/l/linux-gcp/linux-gcp-headers-4.15.0-1008_4.15.0-1008.8_amd64.deb"]
```
as you can see, it is present on test-infra.
So, there is no need for this PR. We just need #826 to be merged.
Moreover, you'll also need the next Falco release (ie: 0.33, to be released end of september) to be able to correctly download these new drivers, because driver-loader was not catching ubuntu flavors before; see: https://github.com/falcosecurity/falco/pull/2178

In the end, i think this PR can be closed. @IanRobertson-wpe wdyt?

0xIAN:
@FedeDP If falco-driver-loader and kernel-crawler sync up with the right name with the updates in 0.33, I think that makes sense, with one caveat. I am doing a deploy to roughly 10,000 GCP hosts this month, and I either need a subset of the eBPF drivers ready now or I have to set up my own build and distribution for those files. So, let's do this. I'll cut down this list of additional driver configs to just those I'll need to get me by for September, and hopefully that will allow this to sail through. Sound like a plan? I'll submit an update to this PR later today, and then I'll just need someone to approve so we can run the test.

FedeDP:
Yeah makes sense. Unfortunately new `ubuntu-$flavor` drivers can only be picked up by next Falco release (because of falco-driver-loader updates).
Therefore, if you really need those drivers now, then this PR (with ubuntu-generic as a target) is the way to go.  
Of course the best and most elegant solution would be the one used by Falco 0.33, ie: using correct target for ubuntu flavors. But this is something that cannot be ported to older Falco versions.

IanRobertson-wpe:
@FedeDP It looks like it still failed. None of those failures involve the kernels I added. The reduction of the configs I added didn't help us fall under the 24 hour runtime threshold either. So, whatever happened here doesn't seem to be a result of the additions I made. Can we please extend the runtime --at least for now -- and get the bad configs removed so I can at least get my changes in and working? I need this to deploy to a very large number of hosts this month.

FedeDP:
I think we could either bump the timeout for now or split the prow jobs, as @zuc suggested.
I am a bit unsure about how to split the `ubuntu-generic-4` jobs, i mean we are already filtering for target (ubuntu-generic) and maj kernel version (4.X). 
Any idea? @leogr @zuc 
Perhaps split it into `4.1*, 4.2*, 4.3*, 4.4*` ... and so on?

Note moreover that we are building bot x86_64 and aarch64 drivers altogether. In the future, we will surely split them in multiple jobs and that would help.
We cannot do it right now because we are not able to filter for architectures, given that we still have to support old driver tree (ie: the one without the $arch in the path). I've got a local branch with these changes however we still need to wait for 2 driverversions in test-infra.

maxgio92:
> Actually I’m more inclined to see if we can either split the job into multiple sub-jobs and/or verify if all the libs driver versions that we are building are still needed.
> 
> Rationale for this is that tweaking the timeout param will result in a blocker for us to evaluate other CI alternatives in the future (e.g. Github Actions), as typically this is not something you can do on hosted solutions.
> 
> What do other @falcosecurity/test-infra-maintainers think?

I think we can in time order consider these options:
1. Extend the jobs timeout
2. Split jobs and restore the jobs timeout

FedeDP:
Hi @0xIAN ! We built and pushed the new drivers for these configs.
I'll trigger a retest now that should skip everything already built, and work.
/test build-new-drivers-ubuntu-generic-4-presubmit

FedeDP:
Yay everything passed!!
/lgtm

@zuc

zuc (approved):
I'm going to approve this but these files will get overridden as soon as the kernel-crawler runs again.

We probably need to better define contribution policies and/or how to deal with overrides @FedeDP @maxgio92

PR #815, opened 2022-08-30
update(config/org): add core-maintainers team

leogr:
Signed-off-by: Leonardo Grasso <me@leonardograsso.com>

### DIFF
--- a/config/org.yaml
+++ b/config/org.yaml
@@ -220,7 +220,7 @@ orgs:
         privacy: closed
         repos:
           advocacy: admin
-      core-maitainers:
+      core-maintainers:
         description: Core maintainers of The Falco Project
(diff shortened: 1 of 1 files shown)


config/org.yaml
```diff
@@ -220,6 +220,23 @@ orgs:
         privacy: closed
         repos:
           advocacy: admin
+      core-maitainers:
```
zuc, line 223:
```suggestion
      core-maintainers:
```

zuc (changes requested):

config/org.yaml
leogr: Thank you!

zuc (approved):

PR #817, opened 2022-08-30
fix(config,driverkit,images): fix kernel-crawler -> test-infra dbg automation

FedeDP:
Current `update-dbg` job is using `run_if_changed` key that is not available for postsubmits job (https://github.com/kubernetes/test-infra/blob/master/prow/jobs.md), thus it is not triggered.
To workaround this issue, we let `update-kernels` job open a PR targeting `kernels` branch on kernel-crawler; then, `update-dbg` will run for merge to that branch.

kernel-crawler `kernels` branch will only have arch subfolders with their lists, and will be protected.

### DIFF
--- a/config/jobs/update-dbg/update-dbg.yaml
+++ b/config/jobs/update-dbg/update-dbg.yaml
@@ -2,9 +2,8 @@ postsubmits:
   falcosecurity/kernel-crawler:
     - name: update-dbg
       decorate: true
-      run_if_changed: '^kernels/'
       branches:
-        - ^main$
+        - ^kernels$
       extra_refs:
       # Check out the falcosecurity/test-infra repo
       # This will be the working directory
--- a/driverkit/utils/scrape_and_generate
+++ b/driverkit/utils/scrape_and_generate
@@ -5,7 +5,7 @@ set -euo pipefail
 DRY_RUN="${DRY_RUN:-false}"
 TMPDIR="$(mktemp -d)"
 ARCH="$1"
-LIST_URL="https://raw.githubusercontent.com/falcosecurity/kernel-crawler/main/kernels/${ARCH}/list.json"
+LIST_URL="https://raw.githubusercontent.com/falcosecurity/kernel-crawler/kernels/${ARCH}/list.json"
 
 function pretty_echo() {
 	echo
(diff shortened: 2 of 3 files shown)


config/jobs/update-dbg/update-dbg.yaml
```diff
@@ -2,9 +2,8 @@ postsubmits:
   falcosecurity/kernel-crawler:
     - name: update-dbg
       decorate: true
-      run_if_changed: '^kernels/'
       branches:
-        - ^main$
+        - ^kernels$
```
FedeDP, line 6: New branch on kernel-crawler with just the kernel lists.

driverkit/utils/scrape_and_generate
```diff
@@ -5,7 +5,7 @@ set -euo pipefail
 DRY_RUN="${DRY_RUN:-false}"
 TMPDIR="$(mktemp -d)"
 ARCH="$1"
-LIST_URL="https://raw.githubusercontent.com/falcosecurity/kernel-crawler/main/kernels/${ARCH}/list.json"
+LIST_URL="https://raw.githubusercontent.com/falcosecurity/kernel-crawler/kernels/${ARCH}/list.json"
```
FedeDP, line 8: The branch is now named `kernels` and will only have the `${ARCH}` subfolders.

images/update-kernels/entrypoint.sh
```diff
@@ -84,13 +85,13 @@ create_pr() {
...
 
-    echo "> creating pull-request to merge ${user}:${branch} into main..." >&2
+    echo "> creating pull-request to merge ${user}:${branch} into ${GH_TARGET_BRANCH} branch." >&2
     body="This PR updates the list of kernels from the latest crawling. Do not edit this PR."
 
     pr-creator \
         --github-endpoint="${GH_PROXY}" \
         --github-token-path="$1" \
-        --org="${GH_ORG}" --repo="${GH_REPO}" --branch=main \
+        --org="${GH_ORG}" --repo="${GH_REPO}" --branch=${GH_TARGET_BRANCH} \
```
FedeDP, line 94: Create a PR from `update-kernels` branch targeting `kernels` branch.

maxgio92:
Good catch @FedeDP! Thanks.

I even prefer this setup, having a dedicated branch for generated arrifacts in kernel-crawler.

maxgio92 (approved):
/approve

FedeDP:
It seems like this was not enough to fix the issue :/

PR #827, opened 2022-09-02
fix(config/org.yaml): remove empty fields

leogr:
See https://prow.falco.org/log?job=peribolos-post-submit&id=1565594745181310976

Signed-off-by: Leonardo Grasso <me@leonardograsso.com>


maxgio92 (approved):
/approve

PR #861, opened 2022-09-23
cleanup(driverkit): dropped a rocky config whose name was wrong

FedeDP:
And whose correct name config is already in place.


maxgio92 (approved):
/approve

PR #875, opened 2022-10-18
Drivers website

Issif:
Signed-off-by: Thomas Labarussias <issif+github@gadz.org>

To allow people to easy search among available drivers, I propose to upload into the [S3 bucket](https://download.falco.org/?prefix=driver/) a website which looks like:

image

To work, we need to generate json files listing all drivers with their details, this is done by a go script. This go script should be run after each merge into /driverkit/config.

### DIFF
--- a/config/jobs/update-drivers-website/update-drivers-website.yaml
+++ b/config/jobs/update-drivers-website/update-drivers-website.yaml
@@ -1,0 +1,23 @@
+periodics:
+  - name: update-drivers-website
+    cron: '0 0 * * *'  # every midnight, daily
+    decorate: true
+    extra_refs:
+    # Check out the falcosecurity/test-infra repo
+    # This will be the working directory
+    - org: falcosecurity
+      repo: test-infra
+      base_ref: master
+      workdir: true
+    spec:
+      containers:
+      # See images/update-drivers-site
+      - image: 292999226676.dkr.ecr.eu-west-1.amazonaws.com/test-infra/update-drivers-website
+        imagePullPolicy: Always
+        command:
+        - /entrypoint.sh
+        env:
+        - name: AWS_REGION
+          value: eu-west-1
+      nodeSelector:
+        Archtype: "x86"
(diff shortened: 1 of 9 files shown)


Issif:
The index.html and the .go are ready, we'll work with @FedeDP to integrate the generation of json files with prow.

images/update-drivers-website/Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM golang:1.18
+
+ADD updateDriversWebsite.go .
+RUN go build -v -o update-drivers-website ./updateDriversWebsite.go
+COPY update-drviers-website /bin
```
FedeDP, line 5: Small typo :)

config/jobs/update-drivers-website/update-drivers-website.yaml
```diff
@@ -0,0 +1,23 @@
+periodics:
+  - name: update-drivers-website
+    cron: '0 0 * * *'  # every midnight, daily
```
maxgio92, line 3: I think this is fine. Just as a note: it means the index and website can have a delay of 23:59:59:... Am I missing something?

If I'm right, we could then find a way to trigger this when the bucket content changes (aka when new artifacts are actually published).

@Issif @FedeDP WDYT?

Issif: We discussed with @FedeDP about that point. We can't trigger the update after each change:
* it will trigger the job a lot of times when several drivers are changed
* we need to wait til the drivers are built and pushed, or it's useless
* the trigger of the jobs to build the drivers doesn't ensure the drivers will be on S3
So we decided the best way is to do a periodic update, we could increase the frequency to be sure.

maxgio92: Totally agree @Issif. Thanks

FedeDP:
Uh it's already merged...

PR #877, opened 2022-10-21
fix(images): fixed update-drivers-website image build and run.

FedeDP:
I am still unsure about whether we will have proper `aws` secrets to push on s3 @maxgio92

### DIFF
--- a/images/update-drivers-website/entrypoint.sh
+++ b/images/update-drivers-website/entrypoint.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 #
-# Copyright (C) 2021 The Falco Authors.
+# Copyright (C) 2022 The Falco Authors.
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


FedeDP:
Ok, i think we need to add:

```
serviceAccountName: driver-kit
```
to the job config file.
See: https://github.com/falcosecurity/test-infra/blob/master/config/prow/build-drivers-serviceaccount.yaml

images/update-drivers-website/entrypoint.sh
```diff
@@ -31,14 +31,14 @@ function check_program {
...
+    check_program "aws"
     check_program "update-drivers-website"
     
     # Build updated json
-    update-drivers-website /tmp/website
+    update-drivers-website images/update-drivers-website/
     
     # Push the updated files to s3 bucket
-    aws s3 cp "/tmp/website/" s3://${S3_DRIVERS_BUCKET}/${S3_DRIVERS_KEY_PREFIX}/site --recursive --include "*.json" --include "index.html" --acl public-read
+    aws s3 cp "images/update-drivers-website/" s3://${S3_DRIVERS_BUCKET}/${S3_DRIVERS_KEY_PREFIX}/site --recursive --include "*.json" --include "index.html" --acl public-read
```
maxgio92, line 41: Instead, why not shipping the index.html artifact with the image, copying it from that and avoid decorating the prowjob with the repostiory? (if the index.html artifact is supposed to be versioned).

In any case, if it makes sense to you @FedeDP, is only an improvement.

maxgio92 (approved):
/approve

PR #878, opened 2022-10-21
encode URI for json + add a .gitignore for locale devs

Issif:
Signed-off-by: Thomas Labarussias <issif+github@gadz.org>

Encode the URI for drivers with a `+` in their names
Add a .gitignore for locale devs

PR #879, opened 2022-10-21
encode only a component

Issif:
Signed-off-by: Thomas Labarussias <issif+github@gadz.org>

enforce the encoding of `lib` only. should fix the #878


maxgio92 (approved):
/approve

PR #881, opened 2022-10-24
fix(dbg): avoid passing a huge kernelconfigdata (see: flatcar) as env variable

FedeDP:
Instead, pass it through a file.
Env variables have size limit of 32KB.
Moreover, dropped replacement of ubuntu-gke with ubuntu-generic now that we support ubuntu-gke.

### DIFF
--- a/driverkit/utils/scrape_and_generate
+++ b/driverkit/utils/scrape_and_generate
@@ -17,9 +17,6 @@ function get_kernel_releases() {
 }
 
 function generate_from_kernel_releases() {
-	local target_match="ubuntu-gke"
-	local target_replace="ubuntu-generic"
-
 	for distro_family in $(jq -cr '. | keys[]' $TMPDIR/sample.json); do
 		while read -r release
 		do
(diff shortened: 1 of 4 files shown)


driverkit/utils/scrape_and_generate
```diff
@@ -29,14 +26,15 @@ function generate_from_kernel_releases() {
...
 			test $DRY_RUN == "true" || \
 				make -C "$(dirname $0)/.." \
 				-e TARGET_ARCH="${ARCH}" \
 				-e TARGET_KERNEL_RELEASE="${release}" \
 				-e TARGET_KERNEL_VERSION="${version}" \
-				-e TARGET_DISTRO="${target/${target_match}/${target_replace}}" \
+				-e TARGET_DISTRO="${target}" \
 				-e TARGET_HEADERS="${kernelurls}" \
-				-e TARGET_KERNEL_DEFCONFIG="${kerneldefconfig}" \
+				-e TARGET_KERNEL_DEFCONFIG="$TMPDIR/kernelconfigdata" \
```
maxgio92, line 37: Thanks @FedeDP. Could you also document this change on `TARGTE_KERNEL_DEFCONFIG` expected value?

FedeDP: Uh you are right sorry!

FedeDP: Mmmh `TARGTE_KERNEL_DEFCONFIG` is not a documented var because we only document filters vars!

maxgio92: You're right too :-) Let's use the opportunity, WDYT? I think the variables specific to `generate` target are missing.

PR #883, opened 2022-10-26
new: add an update-syscalls job that runs monthly

FedeDP:
It will update the supported syscalls by falcosecurity/libs to latest available in the linux kernel, as generic events.

For now it uses `fededp/syscalls-bumper`, but the repo is in the process of being donated to falcosecurity: https://github.com/falcosecurity/evolution/issues/209
I will update the PR accordingly as soon as it is donated.
/hold

### DIFF
--- a/images/update-syscalls/Dockerfile
+++ b/images/update-syscalls/Dockerfile
@@ -1,0 +1,20 @@
+FROM golang:1.18 AS pullrequestcreator
+
+RUN git clone https://github.com/kubernetes/test-infra
+RUN cd test-infra/robots/pr-creator && go build -v -o pr-creator ./main.go
+
+FROM golang:1.19 AS builder
+
+RUN git clone https://github.com/FedeDP/SyscallsBumper.git
+RUN cd SyscallsBumper && make build
+
+FROM debian:buster
+
+COPY --from=pullrequestcreator /go/test-infra/robots/pr-creator/pr-creator /bin
+COPY --from=builder /go/SyscallsBumper/build/bumper /bin
+
+COPY entrypoint.sh /
+
+RUN chmod +x /entrypoint.sh
+
+ENTRYPOINT ["/entrypoint.sh"]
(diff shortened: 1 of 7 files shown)


images/update-syscalls/Dockerfile
```diff
@@ -0,0 +1,20 @@
+FROM golang:1.18 AS pullrequestcreator
+
+RUN git clone https://github.com/kubernetes/test-infra
+RUN cd test-infra/robots/pr-creator && go build -v -o pr-creator ./main.go
+
+FROM golang:1.19 AS builder
+
+RUN git clone https://github.com/FedeDP/syscalls-bumper.git
```
FedeDP, line 8: This will become `falcosecurity/syscalls-bumper` once the donation is done :)

FedeDP:
Imho we can also merge this as-is and then update it to use falcosecurity instead of fededp gh user, once the repo is donated; in this case, we will be able to soon test if the job works correctly (given that it runs on first day of each month!).
/cc @zuc @leogr @maxgio92 

Note that on first run, i don't expect any change to libs because we already have a generated PR open by me: https://github.com/falcosecurity/libs/pull/685

maxgio92:
I agree @FedeDP, I'd be pragmatist :-)

images/update-syscalls/Dockerfile
```diff
@@ -0,0 +1,20 @@
...
+RUN git clone https://github.com/kubernetes/test-infra
+RUN cd test-infra/robots/pr-creator && go build -v -o pr-creator ./main.go
+
+FROM debian:buster
+
+ADD https://github.com/FedeDP/syscalls-bumper/releases/download/v0.0.3/syscalls-bumper_0.0.3_linux_amd64.tar.gz .
+RUN tar -xvf syscalls-bumper_0.0.3_linux_amd64.tar.gz \
+    && chmod +x syscalls-bumper \
+    && mv syscalls-bumper /bin/syscalls-bumper
+
```
maxgio92, line 12: @FedeDP double empty line

FedeDP: Done!

maxgio92 (approved):
/approve

PR #884, opened 2022-10-27
chore(images): dropped useless return which is not error-checked.

FedeDP:


maxgio92 (approved):
/approve

PR #894, opened 2022-11-04
fix: support plugin names containing dashes

jasondellaluce:
For example, this prevents the new `k8saudit-eks` plugin from being released as table


leogr (approved):
Good catch!

/approve

PR #914, opened 2022-12-01
new(driverkit): validate only presubmit build drivers jobs

cappellinsamuele:
Since our dbg is best-effort, there is no need to spend time in building drivers in presubmit jobs since they won't be uploaded anywhere.
Just validate them.

### DIFF
--- a/config/jobs/build-drivers/build-drivers.local
+++ b/config/jobs/build-drivers/build-drivers.local
@@ -14,6 +14,8 @@ presubmits:
         env:
         - name: AWS_REGION
           value: eu-west-1
+        - name: DBG_MAKE_BUILD_TARGET
+          value: validate
         image: localhost:5000/build-drivers
         imagePullPolicy: Always
         securityContext:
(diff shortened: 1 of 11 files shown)


config/jobs/build-drivers/build-drivers.local
```diff
@@ -14,6 +14,8 @@ presubmits:
         env:
         - name: AWS_REGION
           value: eu-west-1
+        - name: DBG_MAKE_BUILD_TARGET
```
FedeDP, line 17: I'd also rename the jobs names since `build-drivers...` is wrong now :D

maxgio92:
Hey @cappellinsamuele! Thank you and welcome! 🤗 

This LGTM, and I agree with you @FedeDP.
Even though I see this validation to be at configuration's syntax-level (with correct configs we don't know if kernel headers packages will be there for the configured kernel release e.g.) I see it as a good trade-off for the same motivation you highlighted @FedeDP.

maxgio92 (approved):
/approve

PR #916, opened 2022-12-02
new(driverkit): configs regenerated after each dbg update

cappellinsamuele:
Every time the dbg gets updated, config files are deleted and generated again in order to always get fresh configurations. In this way there will not be attempts to build old (unbuildable) versions.
Old and already builded drivers will still remain in the bucket.
An alternative would be to delete only configurations older than six months (as in #917 ).

### DIFF
--- a/driverkit/Makefile
+++ b/driverkit/Makefile
@@ -64,7 +64,9 @@ generate:
 		)\
 	)
 
+# clean old configurations before generating the new ones
 generate/auto:
+	rm -rf config/*
 	$(foreach ARCH,$(TARGET_ARCHS),\
 		utils/scrape_and_generate $(ARCH); \
 	)
--- a/driverkit/README.md
+++ b/driverkit/README.md
@@ -131,6 +131,8 @@ You can also validate the configurations you generate with `make validate`, and 
 
 You can find more info about the Driverkit Build Grid [here](#driverkit-build-grid).
 
+It should be noted that the Driverkit Build Grid configurations are kept only for the last kernel-crawler's result, as the crawler represents the uniqe source of truth. Therefore, added configurations will be dropped on Driverkit Build Grid updates but published artifacts will not be cleaned up and will still remain available.
+
 ### Q: How do you publish new drivers?
 
 A: If you have proper S3 permissions from Terraform or Prow, run


driverkit/Makefile
```diff
@@ -65,8 +65,15 @@ generate:
...
+# clean old configurations that could not be buildable anymore
 	$(foreach ARCH,$(TARGET_ARCHS),\
-		utils/scrape_and_generate $(ARCH); \
+		find config/*/${ARCH} -mindepth 1 -type f -delete; \
+	)\
+	
+	find config/*/ -mindepth 1 -type f -delete
+
+	$(foreach ARCH,$(TARGET_ARCHS),\
+		#utils/scrape_and_generate $(ARCH); \
```
FedeDP, line 76: Uh why is this commented? :)

cappellinsamuele: You are right, corrected :)

FedeDP:
/cc @zuc @maxgio92 
I think this is the best approach: 
* we are strongly relying on the automatic flow right now
* even if an user-added config generated artifacts, they will be stored on download.falco.org, even if the config gets later cut by kernel-crawler
* if we aren't able to fetch some kernels, it must be filled as a bug on kernel-crawler

WDYT?
Some numbers: right now we have:
* on x86_64 -> we currently have 7337; by keeping only the ones fetched by latest kernel-crawler run, we would have 6125
* on aarch64 -> we currently have 3222; by keeping only the kernel-crawler ones: 2396

It is a big saving IMHO ;) 

PS: i prefer this approach to the #917 one !

FedeDP:
I am ok with these changes: perhaps we could state somewhere in the docs that our source of truth is the kernel-crawler, and that every config not discovered by it should be considered a bug on it, and thus reported on its repo?
/cc @maxgio92

driverkit/Makefile
```diff
@@ -65,6 +65,13 @@ generate:
 	)
 
 generate/auto:
+# clean old configurations that could not be buildable anymore
+	$(foreach ARCH,$(TARGET_ARCHS),\
```
FedeDP, line 69: I think we might just `rm -rf ${CONFIGS}`.

FedeDP: Ah too many files passed to `rm` :laughing:

maxgio92:
Hi all, doesn't this means @cappellinsamuele @FedeDP that configurations contributed by the community and not automatically generated are cleaned up by the CI this way, does it?
Thanks

FedeDP:
Yes; we might want to state that somewhere perhaps?
Note however that artifacts generated by contributors' configs will stay uploaded.
Fact is, we cannot afford to keep configs for old unsupported kernels around; IMHO any config that is not found should be treated as a kernel-crawler bug.
A contributor can still add a config to generate artifacts for her/his needs, yet our reference truth is the one discovered by kernel-crawler.

maxgio92:
Got it @FedeDP, thank you @cappellinsamuele.
As suggested I'd document this change and highlight that the crawler is the source of truth. Optionally contributors can add configurations - they're encouraged to improve the crawler logics, if possible, as of above - but DBG configurations (as for this PR) are kept only for last crawler's result.
In any case published artifacts are not cleaned up.

driverkit/README.md
```diff
@@ -131,6 +131,8 @@ You can also validate the configurations you generate with `make validate`, and
 
 You can find more info about the Driverkit Build Grid [here](#driverkit-build-grid).
 
+It should be noted that the Driverkit Build Grid configurations are kept only for the last kernel-crawler's result, as the crawler represents the uniqe source of truth. Therefore, added configurations will be dropped on Driverkit Build Grid updates but published artifacts will not be cleaned up and will still remain available.
```
FedeDP, line 134: @maxgio92 here is the note about the source of truth! What do you think?

maxgio92: I think it's good @FedeDP @cappellinsamuele. Thank you!

FedeDP: :love_you_gesture:

cappellinsamuele: Thanks to you!

FedeDP (approved):
/approve

PR #917, opened 2022-12-02
wip: new(driverkit): deletion of older than six months configs on dbg update

cappellinsamuele:
Every time the dbg gets updated, config files that are older than 180 days are deleted to reduce the number of unbuildable configurations. 
Old and already builded drivers will still remain in the bucket.

### DIFF
--- a/driverkit/Makefile
+++ b/driverkit/Makefile
@@ -65,8 +65,15 @@ generate:
 	)
 
 generate/auto:
+# clean old configurations that could not be buildable anymore
 	$(foreach ARCH,$(TARGET_ARCHS),\
-		utils/scrape_and_generate $(ARCH); \
+		find config/*/${ARCH} -mindepth 1 -mtime +180 -type f -delete; \
+	)\
+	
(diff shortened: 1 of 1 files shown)


leogr:
Is it possible to check if the prebuilt driver is actually in the bucket?
If not, it would be a signal that something went wrong, and we should fix it :)

FedeDP:
> If not, it would be a signal that something went wrong, and we should fix it :)

I am not actually sure: using a simple `aws s3api head-object --bucket ${S3_DRIVERS_BUCKET} --key ${S3_DRIVERS_KEY_PREFIX}/${driver_version}/${1}` allows us to test if a certain config had its artifacts built for currently supported driverversions  and we know nothing about other driverversions.
So, if a config actually generated an artifact let's say for driverversion `b7eb0dd65226a8dc254d228c8d950d07bf3521d2`, we cannot check it (unless we do a recursive ls on the bucket, but it would be a bit too expensive!).

No warning is better than wrong warning IMHO :D

FedeDP:
Since it was decided that #916 is the way to go, i think we can close this one, right?

PR #929, opened 2022-12-19
new(config): added build-drivers jobs for bottlerocket.

FedeDP:

### DIFF
--- a/config/jobs/build-drivers/build-new-bottlerocket.yaml
+++ b/config/jobs/build-drivers/build-new-bottlerocket.yaml
@@ -1,0 +1,120 @@
+presubmits:
+  falcosecurity/test-infra:
+  - name: validate-new-drivers-bottlerocket-4-presubmit
+    decorate: true
+    skip_report: false
+    agent: kubernetes
+    branches:
+      - ^master$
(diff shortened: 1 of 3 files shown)


config/jobs/build-drivers/build-new-bottlerocket.yaml
```diff
@@ -0,0 +1,120 @@
+presubmits:
+  falcosecurity/test-infra:
+  - name: validate-new-drivers-bottlerocket-4-presubmit
```
FedeDP, line 3: We only have 4.X and 5.X kernels on bottlerocket.

FedeDP:
Imho btw, i'd completely drop `rebuild-all-drivers` jobs, since they are useless in my opinion.
Moreover, i don't get why there is a presubmit in the `rebuild-all-drivers`, that triggers a publish :/

FedeDP:
Note: we still need to bump driverkit to a version that is capable of building bottlerocket target ;)

FedeDP:
> Note: we still need to bump driverkit to a version that is capable of building bottlerocket target ;)

Done ;)

maxgio92 (approved):
/approve

PR #930, opened 2022-12-19
cleanup(config): dropped rebuild-all-drivers jobs.

FedeDP:
This is just to start a discussion.
Do we still need them?


maxgio92 (approved):
/approve

PR #937, opened 2022-12-22
new(driverkit): support for for native build of drivers on arm64

cappellinsamuele:
This PR:

- Adds the configurations for the jobs in order for them to be scheduled on arm64 nodes when needed
- Drops support for older configs (that do not use semver)
- Removes the unsupported driver version that will be replaced by the new one once released, keeping only the last two. (In this way, for a short time that is until the next release, only two versions will be supported).


FedeDP:
This is neat!! Thank you very much!

I'll hold this one to start a discussion around it (ie: since it drops last non-semver driver version, but we still haven't got the new `4.0.0+driver` driverversion for Falco 0.34, we might stay with only 2 supported driverversions for a couple of weeks; IMHO this is ok, so that we get proper time to test the native arm64 build).

/hold

BTW I think you wanted to squash some commits together since 8fa9554a4df4bc57de68c98ac0bf35a6140091ea has 3 commit messages ;) @cappellinsamuele

cappellinsamuele:
> BTW I think you wanted to squash some commits together since [8fa9554](https://github.com/falcosecurity/test-infra/commit/8fa9554a4df4bc57de68c98ac0bf35a6140091ea) has 3 commit messages ;) @cappellinsamuele
Uh, thank you :D Updated

config/jobs/build-drivers/build-new-amazonlinux.yaml
```diff
@@ -86,13 +86,46 @@ postsubmits:
...
+    skip_report: false
+    agent: kubernetes
+    branches:
+      - ^master$
+    run_if_changed: '^driverkit/config/[a-z0-9.+-]{5,}/aarch64/amazonlinux_*'
+    spec:
+      serviceAccountName: driver-kit
+      containers:
+      - command:
+        - /workspace/build-drivers.sh
```
FedeDP, line 100: I think we should also pass an additional env to `build-drivers.sh` to set the TARGET_ARCH (otherwise we end up trying to build all drivers on each architecture).

Note that `build-drivers.sh` already has a `TARGET_ARCH` (that is 4th argument passed); i think we could just default it to output of `uname -m` (ie: architecture we are currently running on), when it is not passed from cmdline; this will help us avoid touching all config jobs once again!
Basically, we should update this line: https://github.com/falcosecurity/test-infra/blob/86fc92a0816cf53c474fe0cf4801543554a2b4c7/images/build-drivers/build-drivers.sh#L21 adding the default to `$(uname -m)`.

@maxgio92 do you agree? I think it might be a good behavior to default to current architecture.

driverkit/Makefile
```diff
@@ -33,25 +33,11 @@ DRIVERKIT_BACKEND ?= docker
...
 
-specific_target: specific_target_old specific_target_new
-# old fmt had x86_64 only in root folder. Note: this breaks filtering for TARGET_ARCH
-# because specific_target_old will always build any old config.
-# Not a big deal because we were not going to filter for TARGET_ARCH in any case.
-# TODO: rm once old format support is dropped
-specific_target_old: $(patsubst config_%,%,$(subst /,_,$(wildcard config/${TARGET_VERSION}*/${TARGET_DISTRO}_${TARGET_KERNEL}.yaml)))
-# new fmt has $arch subfolders
-specific_target_new: $(patsubst config_%,%,$(subst /,_,$(wildcard config/${TARGET_VERSION}*/${TARGET_ARCH}/${TARGET_DISTRO}_${TARGET_KERNEL}.yaml)))
+specific_target: $(patsubst config_%,%,$(subst /,_,$(wildcard config/${TARGET_VERSION}*/${TARGET_ARCH}/${TARGET_DISTRO}_${TARGET_KERNEL}.yaml)))
```
maxgio92, line 40: Finally 😍 @cappellinsamuele @FedeDP

config/jobs/build-drivers/build-new-amazonlinux.yaml
maxgio92: @FedeDP got it. As of now by default, the wrapper script runs the DBG for all architectures (i.e. `$4` is empty), and I think it should remain as so.
Instead, I would explicitly pass the architecture as argument, as already done with the Linux distrubution (i.e. `amazonlinux`). As the argument accepted as of now is onyl positional, I would introduce the setting of the default value from the environment.
WDYT?

FedeDP: While that would also work, i think that we generally (so, by default) might want to only build drivers for the platform we are currently running on, right?
I mean, it seems a sensible default! But i haven't got a strong opinion about this :) fully open to discuss!

maxgio92: It's not a big deal, we can go on with this PR @FedeDP

FedeDP: Yep the main concern here is that defaulting to "build configs for any architecture" it's gonna work only on x86 nodes, where qemu is supported.
Imo it makes sense to instead default to "build just your own arch", but again either way is fine ahha

FedeDP: Btw i'd give this a try so that next time kernel-crawler runs we will test native arm64 builds.
We can then smooth the rough edges ;)

FedeDP (approved):
Let's test this out with latest crawled kernels!
/approve

FedeDP:
THìhank you very much for this PR! LGTM!

PR #943, opened 2023-01-10
new(build): add job to update the main artifact index with rules from the rules repo

LucaGuerra:

I wish to replicate the same process that we have to update the index file with plugin registry data. Therefore, this PR introduces a new job with a new image that should react to the any registry file change in the rules repo and call the index update program. The program exists in the rules repo and can work on the same falcosecurity index file that we use for plugins, plus the rules registry.

### DIFF
--- a/images/update-rules-index/on-registry-changed.sh
+++ b/images/update-rules-index/on-registry-changed.sh
@@ -1,0 +1,143 @@
+#!/usr/bin/env bash
+#
+# Copyright (C) 2022 The Falco Authors.
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#     http://www.apache.org/licenses/LICENSE-2.0
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+GH_PROXY="${GH_PROXY:-"http://ghproxy"}"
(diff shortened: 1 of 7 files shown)


images/build-plugins/on-registry-changed.sh
```diff
@@ -116,7 +116,7 @@ push_index() {
     pushd "$2"
 
     git add index.yaml
-    git commit --message="update(index.yaml): new plugins registry data" --signoff
+    git commit --message="update(index.yaml): new rules registry data" --signoff
```
FedeDP, line 119: I think this is a small copy/paste error, right? :)

images/update-rules-index/on-registry-changed.sh
```diff
@@ -0,0 +1,143 @@
...
+}
+
+# $1: path of the file containing the token
+# $2: path to the local working copy of the index repo
+push_index() {
+    echo "> pushing distribution index..." >&2
+    pushd "$2"
+
+    git add index.yaml
+    git commit --message="update(index.yaml): new plugins registry data" --signoff
```
FedeDP, line 85: And here too :)

FedeDP (commented):
LGTM aside from a couple of minor typos!

images/build-plugins/on-registry-changed.sh
LucaGuerra: :/

images/update-rules-index/on-registry-changed.sh
LucaGuerra: fixed!

LucaGuerra:
Thank you so much for the review @FedeDP ! As per your additional suggestion, I also added `publish-images` and `build-images` jobs

FedeDP (approved):
/approve

PR #944, opened 2023-01-11
update(images): trigger publish for update-rules-index

LucaGuerra:

The image publishing job should start upon any modification to the image, meaning this would be enough :)


FedeDP (approved):
/approve

PR #946, opened 2023-01-11
fix(update-rules-index): add OCI_REPO_PREFIX

LucaGuerra:

As per title, since the index update option requires it.


FedeDP (approved):
/approve

leogr (approved):
/approve

PR #948, opened 2023-01-12
fix(tf): temporarily revert PR#947

LucaGuerra:

As per title. Sadly we use a version of Terraform that is incompatible with `"terraform-aws-modules/iam/aws//modules/iam-github-oidc-provider"` :sob: . I can't fix it straight away and I don't want to leave test-infra in a broken state so I'm going to revert it and think about a solution

### DIFF
--- a/config/clusters/iam.tf
+++ b/config/clusters/iam.tf
@@ -171,48 +171,6 @@ data "aws_iam_policy_document" "build_plugins_s3_access" {
   }
 }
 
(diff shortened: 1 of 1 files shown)


maxgio92:
Yeah unfortunately yes @LucaGuerra. We need to upgrade Terraform version and bump all the dependencies

maxgio92 (approved):
/approve

PR #951, opened 2023-01-16
fix(config/jobs/build-drivers): corrected nodeSelector position

cappellinsamuele:
Since `nodeSelector` directives have been placed under the wrong section of jobs configuration files, they are moved out from `- command` directives in order to be correctly recognized.


FedeDP (approved):
/approve

PR #954, opened 2023-01-17
fix(config): fixed falcosecurity/rules members.

FedeDP:


leogr (approved):
Good catch!

Thank you

/approve

PR #962, opened 2023-01-20
update(tf): upgrade Terraform to 1.3.7

LucaGuerra:
Thanks to @maxgio92 great work we have updated versions of tf modules that are compatible with 1.x. In addition, I ran `terraform plan` and examined the warnings that were generated, which were mostly related to deprecation of S3 fields and similar things. I followed the documentation to use new recommended resources to achieve the same result.

This update is necessary to enable the GitHub OIDC provider.


leogr (approved):
/approve

PR #972, opened 2023-01-26
chore: falcoctl plugins and config

FedeDP:
* Added `milestone` plugin to falcoctl
* Added `test` as required status check to falcoct
/cc @leogr


leogr (approved):
/approve

PR #980, opened 2023-02-03
fix(config): update-dbg job should only run when a `list.json` file is touched

FedeDP:


config/jobs/update-dbg/update-dbg.yaml
```diff
@@ -4,6 +4,7 @@ postsubmits:
       decorate: true
       branches:
         - ^kernels$
+      run_if_changed: 'list.json$'
```
FedeDP, line 7: Kernel-crawler `kernels` branch paths are:
* x86_64/list.json
* aarch64/list.json
* $AnyNewArch/list.json

maxgio92 (approved):
/approve

PR #983, opened 2023-02-06
cleanup(prow): removed build/publish update-kernels job

cappellinsamuele:
This PR removes update-kernels build and publish jobs from prow configuration files.
Job ported to GHA (see [here](https://github.com/falcosecurity/kernel-crawler/actions/workflows/update-kernels.yml)).


FedeDP (approved):
/approve

PR #989, opened 2023-02-06
chore(config/clusters/ecr-iam): removed update-kernels resource.

cappellinsamuele:
This PR carries a little fix on `update-kernels` job removal form ecr config.


FedeDP (approved):
/approve

PR #1005, opened 2023-02-16
update(config/clusters): ec2 cluster scale-out

cappellinsamuele:
This PR increases eks auto scaling groups capacities to horizontally scale the instances in the cluster: this is helpful to improve the cluster performance on the heaviest jobs (e.g., build-drivers).
Used values are experimental.

### DIFF
--- a/config/clusters/TERRAFORM.md
+++ b/config/clusters/TERRAFORM.md
@@ -22,13 +22,13 @@
 | app\_name | n/a | `any` | n/a | yes |
 | app\_namespace | n/a | `any` | n/a | yes |
 | app\_stage | n/a | `any` | n/a | yes |
-| eks\_arm\_worker\_group\_asg\_desired\_capacity | The Autoscaling Group size of the ARM EKS worker group | `number` | `15` | no |
-| eks\_arm\_worker\_group\_asg\_max\_capacity | The Autoscaling Group size of the ARM EKS worker group | `number` | `25` | no |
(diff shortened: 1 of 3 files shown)


config/clusters/eks_variables.tf
FedeDP, line 1: Do we want to bump `eks_arm_worker_group_asg_min_capacity` too?
Otherwise, i'd keep the old `eks_default_worker_group_asg_min_capacity` (ie: 3), since it was fine i think.

I am also wary of touching the desired worker groups capacities.
@maxgio92 wdyt?

config/clusters/TERRAFORM.md
maxgio92, line 1: I wouln't edit default values, instead update only the values on the `prow.tfvars` file.

config/clusters/eks_variables.tf
maxgio92, line 1: Same as bove: I wouln't edit default values, instead update only the values on the `prow.tfvars` file.

maxgio92: I agree @FedeDP as desired values since we have the cluster-autoscaler installed, are ignored by the EKS Terraform module.

 I don't see problems bumping the minimum, but also keeping the current min value of 1.

FedeDP:
Hey @cappellinsamuele ! I will push on top of this one to follow @maxgio92 suggestions, so that we can go on and merge this one!

cappellinsamuele:
> Hey @cappellinsamuele ! I will push on top of this one to follow @maxgio92 suggestions, so that we can go on and merge this one!

Yes, sure!

maxgio92 (approved):
/approve

PR #1026, opened 2023-03-15
chore(config/prow/monitoring: temporarily expose grafana http only

maxgio92:
This PR configures temporarily the exposition of Prow Monitoring stack's Grafana dashboard at:
- http only
- AWS application load balancer DNS name

/cc @jasondellaluce @FedeDP


FedeDP (approved):
/approve
Good workaround until we are able to expose it as https under `monitoring.prow.falco.org`.

PR #1034, opened 2023-03-16
docs(proposals): align Falco Infra Admins with governace

leogr:
Just found an inconsistency in the documentation and fixed it.


FedeDP (approved):
/approve

PR #1070, opened 2023-03-27
new(config,tools): add Github actions runner controller support.

FedeDP:
For reference, see https://github.com/actions/actions-runner-controller/.

Basically, we introduce a new resource `RunnerDeployment` that will manage github self-hosted runners requests from the falcosecurity/falco repository.

/cc @maxgio92 

/hold

### DIFF
--- a/config/prow/runner.yaml
+++ b/config/prow/runner.yaml
@@ -1,0 +1,13 @@
+apiVersion: actions.summerwind.dev/v1alpha1
+kind: RunnerDeployment
+metadata:
+  name: falco-runner-arm64
+spec:
+  replicas: 1
+  template:
+    spec:
+      dockerdWithinRunnerContainer: true
+      image: summerwind/actions-runner-dind
+      repository: falcosecurity/falco
+      nodeSelector:
+        Archtype: "arm"
--- a/tools/deploy_prow.sh
+++ b/tools/deploy_prow.sh
@@ -74,7 +74,7 @@ function launchMonitoring(){
 }
 
 function launchARC(){
-  kubectl apply -f https://github.com/actions/actions-runner-controller releases/download/v0.22.0/actions-runner-controller.yaml
+  kubectl apply -f https://github.com/actions/actions-runner-controller/releases/download/v0.27.1/actions-runner-controller.yaml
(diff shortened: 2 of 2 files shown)


config/prow/runner.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: actions.summerwind.dev/v1alpha1
+kind: RunnerDeployment
+metadata:
+  name: falco-runner-arm64
+spec:
+  replicas: 1
+  template:
+    spec:
+      dockerdWithinRunnerContainer: true
```
FedeDP, line 9: See https://github.com/actions/actions-runner-controller/blob/c23e31123c365f4f2bdee4d8b84b9dd48e246d56/docs/using-entrypoint-features.md?plain=1#L60

tools/deploy_prow.sh
```diff
@@ -71,6 +73,12 @@ function launchMonitoring(){
   make -C config/prow/monitoring
 }
 
+function launchARC(){
+  kubectl apply -f https://github.com/actions/actions-runner-controller releases/download/v0.22.0/actions-runner-controller.yaml
+  kubectl create secret generic controller-manager -n actions-runner-system --from-literal=github_token=${PROW_OAUTH_TOKEN}
```
LucaGuerra, line 78: note to self: double check that the `PROW_OAUTH_TOKEN` is the right variable to use

LucaGuerra: I can confirm it should be the right one

tools/deploy_prow.sh
```diff
@@ -71,6 +73,12 @@ function launchMonitoring(){
   make -C config/prow/monitoring
 }
 
+function launchARC(){
+  kubectl apply -f https://github.com/actions/actions-runner-controller releases/download/v0.22.0/actions-runner-controller.yaml
```
maxgio92, line 77: Typo in the URL (` ` instead of `/` in `controller releases/download/...`).
Also, I'd pick the latest stable [`v0.27.1`](https://github.com/actions/actions-runner-controller/releases/tag/v0.27.1).

FedeDP: Great catch! (copy pasted from their quickstart :laughing: )

FedeDP: Fixed, thank you!

LucaGuerra (approved):

PR #1073, opened 2023-03-28
revert: "Bump prow from v20230315-6d54c174f4 to v20230324-8697ff4d24"

FedeDP:
Reverts falcosecurity/test-infra#1071

Seems like it is causing issues to the prow-controller-manager, since no prow job could be scheduled after that PR was merged: https://prow.falco.org/

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -34,10 +34,10 @@ plank:
       timeout: 24h # Up to 24 hours for driverkit builder jobs
       grace_period: 10m
       utility_images:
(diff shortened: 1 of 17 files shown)


FedeDP:
I think we need an admin to manually merge this, since CI is completely freezed because of the prow cluster issues :1st_place_medal:

maxgio92:
I agree with @FedeDP

maxgio92:
What do you think @falcosecurity/admins?

PR #1076, opened 2023-04-03
adding EXONER4TED to kernel-crawler org

EXONER4TED:
Adding myself to the kernel-crawler org as requested by @FedeDP here: https://github.com/falcosecurity/kernel-crawler/pull/108#issuecomment-1494084818


FedeDP (approved):
/approve

PR #1081, opened 2023-04-04
fix(prow): do not fail if the secret already exists

LucaGuerra:
As per title, follow the same pattern used in the rest of the repo


FedeDP (approved):
/approve

Great catch :tm: !

PR #1082, opened 2023-04-05
fix(config/clusters): added policies for S3 access update-driver-website workflow

cappellinsamuele:
This PR adds the policy to allow update-driver-website workflow to upload to S3

### DIFF
--- a/config/clusters/iam.tf
+++ b/config/clusters/iam.tf
@@ -218,6 +218,44 @@ data "aws_iam_policy_document" "rules_s3_access" {
   }
 }
 
+# Test-infra repository
+
+module "test-infra_s3_role" {
+  source    = "terraform-aws-modules/iam/aws//modules/iam-github-oidc-role"
+  version = "5.10.0"
+  create = true
+  subjects = [
+    "falcosecurity/test-infra:ref:refs/heads/master"
+  ]
+  policies = {
+    test-infra_s3_access = "${aws_iam_policy.test-infra_s3_access.arn}"
+  }
+}
+
+resource "aws_iam_policy" "test-infra_s3_access" {
+  name_prefix = "github_actions-test-infra-s3"
+  description = "GitHub actions S3 access policy for test-infra update-drivers-website workflow"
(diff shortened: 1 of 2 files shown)


config/clusters/iam.tf
```diff
@@ -218,6 +218,44 @@ data "aws_iam_policy_document" "rules_s3_access" {
...
+    effect = "Allow"
+    actions = [
+      "s3:PutObject",
+      "s3:GetObjectAcl",
+      "s3:GetObject",
+      "s3:DeleteObject",
+      "s3:PutObjectAcl"
+    ]
+    resources = [
+      "arn:aws:s3:::falco-distribution/driver/site/*",
```
FedeDP, line 253: We cannot escape `falco-distribution/driver/site` ;) Thus we should not be able to make any damage (eg: deleting drivers)!

config/clusters/iam.tf
```diff
@@ -218,6 +218,44 @@ data "aws_iam_policy_document" "rules_s3_access" {
   }
 }
 
+# Test-infra repository
+
+module "test-infra_s3_role" {
+  source    = "terraform-aws-modules/iam/aws//modules/iam-github-oidc-role"
+  version = "5.10.0"
```
LucaGuerra, line 225:
```suggestion
  version = "5.10.0"
  name = "github_actions-test-infra-s3"
```
Just a nit but if you add this the role name will be predictable!

FedeDP (approved):
/approve

LucaGuerra:
Great job @cappellinsamuele ! :rocket:

config/clusters/iam.tf
```diff
@@ -218,6 +218,45 @@ data "aws_iam_policy_document" "rules_s3_access" {
...
+  name = "github_actions-test-infra-s3"
+  create = true
+  subjects = [
+    "falcosecurity/test-infra:ref:refs/heads/master"
```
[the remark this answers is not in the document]
FedeDP: :pray:

PR #1094, opened 2023-04-12
chore(config/clusters/prow.tfvars): increased asg capacities.

cappellinsamuele:
This PR extends both the `x86` and `arm64` autoscaling groups capacities.


maxgio92 (approved):
/approve

PR #1095, opened 2023-04-12
update(config/jobs/build-drivers): tuning pods resources allocation

cappellinsamuele:
This PR tunes the resource allocation to pods responsible for build-driver jobs.
The requested CPU and memory have been decreased and limits have been set to try to schedule two pods on the same node.
Since build processes are not very memory intensive, pods should not be affected by OOM kills.
On the other hand, requests and limits on the CPU might be not sufficient and cause throttling.
Also, the range between CPU requests and limits may be too small resulting in pods being unable to be scheduled.

WDYT @FedeDP @maxgio92?

### DIFF
--- a/config/jobs/build-drivers/build-new-amazonlinux.yaml
+++ b/config/jobs/build-drivers/build-new-amazonlinux.yaml
@@ -23,9 +23,12 @@ presubmits:
         securityContext:
           privileged: true
         resources:
+          limits:
+            cpu: 1.0
+            memory: 4Gi
           requests:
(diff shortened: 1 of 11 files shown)


cappellinsamuele:
> Thanks @cappellinsamuele. Should be enough the memory limit, maybe 1 CPU is not so good for compilation. Anyway the risk is _only_ throttling.
> 
> /lgtm

I agree that 1 CPU is not so good, but the goal would be to force the scheduling of two pods in a single node, even by sacrificing some build time.
In this way, it should not be necessary to manually restart jobs.

FedeDP (approved):
Let's try this on next monday :)
/approve

PR #1096, opened 2023-04-12
update(config/jobs/build-drivers): added jobs for ubuntu 6.x and centos 6.x kernel versions

### LINKED ISSUE #931: Support build-drivers for linux 6.0
**Motivation**

Most of our build-drivers jobs are split for each kver; but we only support linux up to 5.X.

**Feature**

We should add linux 6.X support.

cappellinsamuele:
This PR introduces the configurations for build-driver jobs of `ubuntu 6.x` and `centos 6.x` kernel versions.

Fixes #931


FedeDP (approved):
/approve

PR #1097, opened 2023-04-13
Add new driver kit config

rao-hassan-1:


rao-hassan-1:
@FedeDP how do I check if they are okay?

FedeDP:
I mean, where did you get them from?

driverkit/config/4.0.0+driver/x86_64/debian_5.10.0-0-amd64_1.yaml
```diff
@@ -0,0 +1,8 @@
+kernelversion: 1
+kernelrelease: 5.10.0-0.deb10.17-amd64
+target: debian
+architecture: amd64
+output:
+  module: output/4.0.0+driver/x86_64/falco_debian_5.10.0-0.deb10.17-amd64_1.ko
+  probe: output/4.0.0+driver/x86_64/falco_debian_5.10.0-0.deb10.17-amd64_1.ko
```
FedeDP, line 7: Output probe must end with `.o`.

FedeDP:
Also, can you add the same config for other driverversions, ie:
* 2.0.0+driver
* 3.0.1+driver
?
So that users using older Falco versions will have the driver too!

rao-hassan-1:
I got the `kernerurls` from this file driverkit/config/4.0.0+driver/x86_64/debian_5.10.103-1-amd64_1.yaml. Should I omit them? 

Here's the output from running the commands 

```
root@cde-qd2mj:/# uname -r
5.10.0-0.deb10.17-amd64
root@cde-qd2mj:/# uname -v
#1 SMP Debian 5.10.136-1~deb10u3 (2022-09-06)
root@cde-qd2mj:/# 
```

driverkit/config/4.0.0+driver/x86_64/debian_5.10.0-0-amd64_1.yaml
rao-hassan-1: Thanks for catching this!

FedeDP:
> I got the kernerurls from this file driverkit/config/4.0.0+driver/x86_64/debian_5.10.103-1-amd64_1.yaml. Should I omit them?

Well fact is, you need the correct headers; in fact they are the most important thing specified in the config file. Otherwise you will be building a driver against a wrong kernel release, and your kernel won't let you inject it.

So, i think they are wrong. Unfortunately, omitting them is not going to work; i just checked and driverkit is not able to fetch the correct headers either (unsurprisingly since it uses the same logic as kernel-crawler that has this bug :/ ).
I think best option is to wait for a fix on kernel-crawler side, or try to find correct headers!

rao-hassan-1:
I see. Do you know how I would be able to find the correct headers? 

I am not on debian buster.

FedeDP:
> Do you know how I would be able to find the correct headers?

Unfortunately no. Indeed that's the reason why we weren't still able to fix aforementioned kernel-crawler issue.
Once we find where are these headers, we can adapt our logic  to crawl for them too!

rao-hassan-1:
Makes sense. Do you know of a workaround that will let me install falco on the K8s cluster provided by Digital Ocean?

FedeDP:
Yep, you can try to use the modern-bpf driver for now: https://github.com/falcosecurity/charts/tree/master/falco#deploying-falco-in-kubernetes

FedeDP:
On a second thought, given your `uname` outputs above, i think you are being hit by https://github.com/falcosecurity/falco/issues/2374.

rao-hassan-1:
The `btf`  directory does not exist in my kernel so I cannot use modern-bpf aprroach

FedeDP:
Since https://github.com/falcosecurity/falco/pull/2377#issuecomment-1506847978, i think we can close this one, right?
Thank you for your time and for your help in testing this!

rao-hassan-1:
Yes. I have closed the PR.

PR #1102, opened 2023-04-19
new(config): added iam s3 access roles for Falco (dev and release).

FedeDP:
I decided to split dev and release so that the master CI is not able to push anything to the release packages.

See https://github.com/falcosecurity/falco/pull/2501

### DIFF
--- a/config/clusters/iam.tf
+++ b/config/clusters/iam.tf
@@ -297,6 +297,84 @@ data "aws_iam_policy_document" "test-infra_s3_access" {
   }
 }
 
+# Falco repository (dev packages)
+
+module "falco_dev_s3_role" {
+  source    = "terraform-aws-modules/iam/aws//modules/iam-github-oidc-role"
+  version = "5.10.0"
+  name = "github_actions-falco-dev-s3"
+  create = true
+  subjects = [
+    "falcosecurity/falco:ref:refs/heads/master"
+  ]
+  policies = {
+    falco_s3_access = "${aws_iam_policy.falco_dev_s3_access.arn}"
+  }
+}
+
+resource "aws_iam_policy" "falco_dev_s3_access" {
+  name_prefix = "github_actions-falco-dev-s3"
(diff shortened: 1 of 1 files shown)


config/clusters/iam.tf
```diff
@@ -297,6 +297,84 @@ data "aws_iam_policy_document" "test-infra_s3_access" {
...
 
+# Falco repository (dev packages)
+
+module "falco_dev_s3_role" {
+  source    = "terraform-aws-modules/iam/aws//modules/iam-github-oidc-role"
+  version = "5.10.0"
+  name = "github_actions-falco-dev-s3"
+  create = true
+  subjects = [
+    "falcosecurity/falco:ref:refs/heads/master"
```
FedeDP, line 308: Dev has only `master` pushes.

config/clusters/iam.tf
```diff
@@ -297,6 +297,84 @@ data "aws_iam_policy_document" "test-infra_s3_access" {
...
+    effect = "Allow"
+    actions = [
+      "s3:PutObject",
+      "s3:GetObjectAcl",
+      "s3:GetObject",
+      "s3:DeleteObject",
+      "s3:PutObjectAcl"
+    ]
+    resources = [
+      "arn:aws:s3:::falco-distribution/packages/*-dev/*",
```
FedeDP, line 333: I am not sure whether this regex is fine, but i think so.
Basically, master workflow can only PUT packages to the `dev` packages buckets.

LucaGuerra: I'll try to check if this can work, I'm not sure either

LucaGuerra: Turns out I didn't have an env set up for this test, but according to the documentation we're fine: https://docs.aws.amazon.com/AmazonS3/latest/userguide/s3-arn-format.html

FedeDP: Yep that's what i read too!

leogr (approved):
/approve

PR #1104, opened 2023-04-19
cleanup(images/build-plugins): build-plugins scripts cleanup

cappellinsamuele:
Followup of #1103.

PR #1103, opened 2023-04-19
cleanup(config/jobs/build-plugins/build-plugins.yaml): removed build-plugins jobs configs

cappellinsamuele:
This PR removes `build-plugins` jobs since they have been ported to [plugins](https://github.com/falcosecurity/plugins) repo (see [#238](https://github.com/falcosecurity/plugins/pull/238)).

PR #1098, opened 2023-04-13
new(config/clusters/iam.tf): added policies for S3 access plugins repo GHA

cappellinsamuele:
This PR adds the policy to allow access to the S3 bucket to GHA in [plugins](https://github.com/falcosecurity/plugins) repo.

PR #1105, opened 2023-04-20
cleanup(config/clusters/iam.tf): removed policy for build-plugins job S3 access.

cappellinsamuele:
This PR completes #1103 and #1104 by removing the `build-plugins` job S3 policy in favor of gha (see #1098).


FedeDP (approved):
/approve

PR #1120, opened 2023-05-02
Add postsubmit job to sync Peribolos GitHub Teams from OWNERS

### LINKED ISSUE #777: Automatically sync `config/org.yaml` when an `OWNERS` file gets updated
**Motivation**

We need to automate the way we manage repository permissions.

On the one hand, `OWNERS` files are honored by Prow. On the other hand, GitHub uses its settings which are enforced by Peribolos that applies settings we commit to [config/org.yaml](https://github.com/falcosecurity/test-infra/blob/master/config/org.yaml).

Unfortunately, there's no automatic mechanism to sync the two. And manually synching them can't scale (see https://github.com/falcosecurity/test-infra/pull/776).

**Feature**

Introduce a mechanism to automatically update [config/org.yaml](https://github.com/falcosecurity/test-infra/blob/master/config/org.yaml) when an `OWNERS` file gets updated.

**Alternatives**

Manual synch. Honestly, it's a pain.

**Additional context**

maxgio92:
The post-submit Prow Job runs on changes on OWNERS files - i.e. in the evolution repository.

It synchronizes the Peribolos config - in the test-infra repository - updating the specified evolution-maintainers GitHub Team, with the additional approvers specified in the OWNERS file.

More [here](https://github.com/maxgio92/peribolos-syncer/blob/v0.1.0/docs/peribolos-syncer_sync_github.md#peribolos-syncer-sync-github).

Closes #777

### DIFF
--- a/config/jobs/update-github-teams/peribolos-syncer.yaml
+++ b/config/jobs/update-github-teams/peribolos-syncer.yaml
@@ -9,7 +9,7 @@ postsubmits:
     run_if_changed: 'OWNERS$'
     spec:
       containers:
-      - image: gcr.io/maxgio92/peribolos-syncer:0.1.0
+      - image: gcr.io/maxgio92/peribolos-syncer:0.2.0
         args:
         - sync
         - github
@@ -20,7 +20,7 @@ postsubmits:
         - --peribolos-config-git-ref=master
         - --owners-repository=evolution
         - --owners-git-ref=main
-        - --owners-path=OWNERS
+        - --approvers-only=true
         - --git-author-name=poiana
         - --git-author-email="51138685+poiana@users.noreply.github.com"
         - --gpg-public-key=/tmp/gpg-pub/poiana.pub
--- a/config/jobs/update-github-teams/peribolos-syncer.yaml
+++ b/config/jobs/update-github-teams/peribolos-syncer.yaml
@@ -1,0 +1,116 @@
+postsubmits:
+  falcosecurity/evolution:
+  - name: peribolos-syncer-post-submit
+    branches:
+    - ^main$
+    decorate: true
+    max_concurrency: 1
+    skip_report: false
+    run_if_changed: 'OWNERS$'
+    spec:
+      containers:
+      - image: gcr.io/maxgio92/peribolos-syncer:0.1.0
+        args:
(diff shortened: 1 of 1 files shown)


maxgio92:
PTAL @leogr. WDYT?

A demo PR is #1117

leogr (commented):
It seems great! I reserve some time for a second look. Meanwhile a question: can we enable it for all repos without the need of copy/paste the job? We use OWNERS the n all repos.

maxgio92:
I don't think so @leogr because the flow is to react to changes to the leaf OWNERS file, when PRs are merged, on the main branch, on the single repositories.

So I think we need anyway one postsubmit Prow job, per repository. WDYT?

leogr:
In my experience copy/paste has been very error prone in this context. However, if there's no alternative, we have to deal with it :)

Further question: does it work with nested OWNERS files? We can have multiple OWNERS in a repo, with some of them in subfolders.

maxgio92:
@leogr I see your point. In order to provide granularity with an event-based pull approach like this with jobs I'm not sure we can deal with it. An alternative - it could be a next version - is by using a global webhook (maybe a Prow plugin as @jasondellaluce suggested).

As of now it doesn't support an OWNERS tree but only a specified one (i.e. the leaf approvers declared in the in the root OWNERS file). Anyway the support can be added easily (the underlying owners package already supports it).

maxgio92:
Question: do you think @leogr in our use case, an approver for a particular sub-part of a repository, should be the member of the related repository's maintainers GitHub team?

leogr:
> Question: do you think @leogr in our use case, an approver for a particular sub-part of a repository, should be the member of the related repository's maintainers GitHub team?

It should, and we are already following this pattern AFAIK

maxgio92:
I agree @leogr. Considering this point I'd keep this PR on hold, until the feature is implemented in the tool.

/hold

maxgio92:
/unhold

Implemented the change @leogr

leogr (commented):
/lgtm

just waiting for a second opinion cc @FedeDP

FedeDP (approved):
LGTM!
Thank you @maxgio92 !
/approve

PR #1122, opened 2023-05-03
fix(iam): add CloudFront permissions to S3 access

LucaGuerra:
CloudFront cannot be a separate role but rather an addition to the S3 one, since S3 updates and CF invaliation happen in the same step.


LucaGuerra:
cc @FedeDP

FedeDP (approved):
/approve

PR #1127, opened 2023-05-08
update(config): remove a no more required job for libs

Andreagit97:
This PR removes a  no more required job on libs, instead of `build-modern-bpf-s390x 😁 (system_deps)`  no we use `build-libs-s390x 😁 (system_deps)`


leogr (approved):
/approve

PR #1134, opened 2023-05-12
feat(aws/eks): add new and move jobs nodes to single-az

maxgio92:
> Note: this will re-create all the node groups, besides adding a new one deployed in single-AZ.

/cc @zuc @FedeDP @cappellinsamuele

### DIFF
--- a/config/clusters/eks.tf
+++ b/config/clusters/eks.tf
@@ -29,8 +29,8 @@ module "eks" {
 
   #Managed Node Group
   node_groups = {
-    falco-ng = {
-      name_prefix        = null
+    default = {
+      name               = var.eks_default_worker_group_name
       desired_capacity   = var.eks_default_worker_group_asg_desired_capacity
       max_capacity       = var.eks_default_worker_group_asg_max_capacity
       min_capacity       = var.eks_default_worker_group_asg_min_capacity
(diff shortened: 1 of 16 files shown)


config/clusters/eks_variables.tf
```diff
@@ -14,7 +16,7 @@ variable "eks_default_worker_group_instance_type" {
 }
 
 variable "eks_default_worker_group_asg_min_capacity" {
-  default     = 3
+  default     = 1
```
maxgio92, line 19: The minimum of one node could be enough to run Prow microservices-only.
It's anyway auto-scaled by the cluster-autoscaler deployed on Kubernetes.

FedeDP (commented):
/lgtm

maxgio92:
Note: right now jobs miss the toleration for the singleaz nodes taint

maxgio92:
Integrated by #1136

maxgio92:
Ready to be reviewed @FedeDP

config/clusters/eks.tf
```diff
@@ -40,29 +40,62 @@ module "eks" {
...
+
+    jobs_arm = {
+      name             = var.eks_jobs_arm_worker_group_name
+      desired_capacity = var.eks_jobs_arm_worker_group_asg_desired_capacity
+      max_capacity     = var.eks_jobs_arm_worker_group_asg_max_capacity
+      min_capacity     = var.eks_jobs_arm_worker_group_asg_min_capacity
+      instance_types   = [var.eks_jobs_arm_worker_group_instance_type]
+
+      ami_type = "AL2_ARM_64"
+      taints   = [local.arm_nodegroup_taint]
```
FedeDP, line 86: Shouldn't this have also the `local.single_az_nodegroup_taint`?

config/clusters/eks_variables.tf
```diff
@@ -30,20 +32,64 @@ variable "eks_default_worker_group_asg_max_capacity" {
...
+  description = "The Autoscaling Group size of the jobs EKS worker group"
+}
+
+variable "eks_jobs_worker_group_asg_desired_capacity" {
+  default     = 4
+  description = "The Autoscaling Group size of the jobs EKS worker group"
+}
+
+variable "eks_jobs_worker_group_asg_max_capacity" {
```
FedeDP, line 60: Shouldn't we upload namings in prow.tfvars too? Ie: https://github.com/falcosecurity/test-infra/blob/master/config/clusters/prow.tfvars#L10C50-L12

config/clusters/eks.tf
maxgio92: Nope @FedeDP. It's deployed in multi-AZ here

config/clusters/eks_variables.tf
maxgio92: Good catch @FedeDP, thank you.

config/clusters/eks.tf
FedeDP: Mmh so build drivers arm64 jobs will still risk to be evicted?

config/clusters/eks_variables.tf
maxgio92: Whoops, sorry, I misunderstood. We don't need it, we have default values right now

config/clusters/eks.tf
maxgio92: I didn't see the amount of eviction for arm builds right now. But we can do it also for them. WDYT?

cappellinsamuele: I can confirm that we experienced some evictions also from arm nodes in one of the last runs, so maybe it's a good idea to add the taint

maxgio92: Applied @FedeDP @cappellinsamuele :)

FedeDP (approved):
/approve

PR #1140, opened 2023-05-15
update(iam): allow dev role for prerelease tags

LucaGuerra:
When this role was set up we could only write to the `-dev` bucket from the master branch in Falco. This has since changed because prerelease support has been added which allow us to use development tags that will ultimately end up in the `-dev` bucket. Meaning that we need to add the appropriate permission here.


FedeDP (approved):
/approve

FedeDP:
Closing and reopening.
/close

PR #1146, opened 2023-05-15
perf(config/clusters): tune node groups max horizontal capacity

### LINKED ISSUE #1145: Increase max capacity of jobs on x86 node group
**Motivation**

Driver building jobs are triggered weekly and the spike could need more than 10 nodes.

**Feature**

Increase the max capacity from 10 to 20.

**Alternatives**

NA

**Additional context**

The current max capacity of the jobs on x86-node group is 10.
The desired size is managed by the cluster-autoscaler, installed on Kubernetes.

Furthermore, as the node group for Prow is now dedicated to its microservices (aka it does no longer run jobs by default), its max capacity can be reduced to 10 (or even less).

maxgio92:
Fixes #1145 

/cc @FedeDP


FedeDP (approved):
/approve

PR #1149, opened 2023-05-15
fix(config/jobs/autobump/autobump.yaml): removed unused ssh key volume mount.

cappellinsamuele:
This PR is needed to fix the [`autobump`](https://github.com/falcosecurity/test-infra/blob/master/config/jobs/autobump/autobump.yaml) Prow Job since it gets stuck in trying to mount an inexistent volume containing an unnecessary ssh key.


maxgio92 (approved):
/approve

Thank you @cappellinsamuele!

PR #1154, opened 2023-05-16
chore(config): fixed some validate-new-drivers job names.

FedeDP:


maxgio92 (approved):
/approve

Good catch, thanks @FedeDP

PR #1165, opened 2023-05-22
refactor(config/clusters): move jobs on arm to different subnet

### LINKED ISSUE #1164: Node group ARM cannot scale-up
**Describe the bug**

Node group ARM cannot scale-up.

**How to reproduce it**

NA.

**Expected behaviour**

To scale-up.

**AWS Logs**

```
Amazon AutoScaling was unable to launch instances because there are not enough free addresses in the subnet associated with your AutoScaling group(s).
```

**Environment**

NA

**Additional context**

NA

maxgio92:
This is a quick solution to avoid horizontal scaling not working on the jobs on ARM's node group.
It needs a better implementation then.

Fixes #1164 

/cc @FedeDP


FedeDP (approved):
/approve

PR #1181, opened 2023-06-12
chore(drivekit/configs): needed to trigger building job for minikube …

alacuku:
chore(drivekit/configs): needed to trigger building job for minikube drivers


FedeDP (approved):
/approve

FedeDP:
We had a bug when we added the `5.0.1+driver` driverversion and not all drivers were built!

PR #1186, opened 2023-06-22
update(config): updated autobump image version and configuration.

cappellinsamuele:
This PR updates the autobumper image and configuration to make Poiana signing its commits (see [#29591](https://github.com/kubernetes/test-infra/pull/29591)) and to exclude from the analysis the `driverkit` path.


FedeDP (approved):
This is epic!
/approve

PR #1189, opened 2023-06-29
chore(config/prow/arc): increment falco arm64 runners replicas.

FedeDP:
This way, we can support: PR CI on arm64 and master/release CI static arm64 binary.

### DIFF
--- a/config/prow/arc/arm-runner.yaml
+++ b/config/prow/arc/arm-runner.yaml
@@ -3,7 +3,7 @@ kind: RunnerDeployment
 metadata:
   name: falco-runner-arm64
 spec:
-  replicas: 1
+  replicas: 5
   template:
     spec:
       dockerdWithinRunnerContainer: true
--- a/config/prow/arc/arm-runner.yaml
+++ b/config/prow/arc/arm-runner.yaml
@@ -3,7 +3,6 @@ kind: RunnerDeployment
 metadata:
   name: falco-runner-arm64
 spec:
-  replicas: 5
   template:
     spec:
       dockerdWithinRunnerContainer: true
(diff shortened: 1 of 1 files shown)


FedeDP:
/cc @maxgio92 

This leaves us with 15 nodes for arm64 drivers building pipeline (that means up to 30 pods). I think it is ok; we could also bump max autoscaling to 25, but i don't think it is worth it now.

FedeDP:
The new implementation uses autoscaling based on the job queue for Falco repository.
/cc @maxgio92

config/prow/arc/arm-runner.yaml
```diff
@@ -21,3 +20,18 @@ spec:
         operator: "Equal"
         value: "SingleAZ"
         effect: "NoSchedule"
+---
```
maxgio92, line 23: Interesting @FedeDP!

maxgio92:
> /cc @maxgio92
> 
> This leaves us with 15 nodes for arm64 drivers building pipeline (that means up to 30 pods). I think it is ok; we could also bump max autoscaling to 25, but i don't think it is worth it now.

I think it's ok @FedeDP.
/lgtm

maxgio92 (approved):
/approve

FedeDP:
Keep this on hold for now to avoid any disruptive change before the Falco release (even if we got plenty of time to revert or fix this :D )

FedeDP:
No more needed since we are now using `actuated.dev` (https://actuated.dev/) provided runners.
/close

FedeDP:
Indeed, once Falco CI is working with new runnes, we can remove the `arc` completely from test-infra.

PR #1190, opened 2023-06-29
fix(autobumper): moved signoff option from config to job args.

cappellinsamuele:


FedeDP (approved):
Great catch :D
/approve

PR #1133, opened 2023-05-11
cleanup(config/clusters/ecr.tf): autobump ecr force delete.

cappellinsamuele:
This PR forces delete `autobump` ecr resource.

PR #1132, opened 2023-05-11
cleanup(images/autobump): old autobump image removal.

cappellinsamuele:
This PR completes #1130 by removing the old and no longer used `autobump` job image.

PR #1194, opened 2023-07-05
cleanup(config/clusters/ecr.tf): removed old autobumper ecr resource.

cappellinsamuele:
Completes #1132 and #1133 .


FedeDP (approved):
/approve

PR #1204, opened 2023-08-02
wip: new(images): added initial work on a `dbg-go` tool.

FedeDP:
Basically, the beginning of replacing our `dbg` tools (see test-infra/driverkit subfolder) and makefiles with a go tool, that will be faster and safer to use.

Location might change but it will be installed in the system inside the `update-dbg` image, and possible the `build-drivers` one too.

Opening this as wip to track it.

It needs to replace:
- [ ] utils/scrape_and_generate
- [ ] utils/generate
- [ ] utils/checkfiles
- [ ] utils/cleanup_s3
- [ ] utils/driverstats
- [ ] utils/build
And all the Makefile's logic.

The first commit starts with implementing an initial stage for `scrape_and_generate` and `generate` to be used by the `update-dbg` image (that is too slow nowadays to build dbg configs!).


images/update-dbg/dbg-go/cmd/root.go
```diff
@@ -0,0 +1,32 @@
...
+		PersistentPreRunE: func(cmd *cobra.Command, args []string) error {
+			return viper.BindPFlags(cmd.Flags())
+		},
+	}
+)
+
+// NewRootCmd instantiates the root command.
+func init() {
+	flags := rootCmd.PersistentFlags()
+	flags.Bool("dry-run", false, "enable dry-run mode.")
```
FedeDP, line 24: We will surely need more common options!

images/update-dbg/dbg-go/pkg/autogenerate/autogenerate.go
```diff
@@ -0,0 +1,121 @@
...
+		}
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+
+	// Fetch kernel list json
+	jsonData, err := utils.GetURL(url)
+	//jsonData, err := os.ReadFile("/home/federico/Scaricati/kernels.json")
```
FedeDP, line 69: TODO: cleanup :)

images/update-dbg/dbg-go/pkg/autogenerate/autogenerate.go
```diff
@@ -0,0 +1,121 @@
...
+				KernelVersion:    kernelEntry.KernelVersion,
+				KernelRelease:    kernelEntry.KernelRelease,
+				Target:           kernelEntry.Target,
+				Architecture:     utils.ToDebArch(opts.Architecture),
+				KernelUrls:       kernelEntry.Headers,
+				KernelConfigData: kernelEntry.KernelConfigData,
+			}
+
+			for _, oDir := range outputDirs {
+				// FIXME! use correct path (inside folder) https://github.com/falcosecurity/test-infra/blob/master/driverkit/utils/generate#L93
```
FedeDP, line 100: We need to fix a couple of things here, then we are good 2 go and we can benchmark this against current bash scripts.

FedeDP: I will share some numbers asap!

images/update-dbg/dbg-go/pkg/autogenerate/types.go
```diff
@@ -0,0 +1,51 @@
...
+	KernelRelease    string   `json:"kernelrelease"`
+	Target           string   `json:"target"`
+	Headers          []string `json:"headers"`
+	KernelConfigData []byte   `json:"kernelconfigdata"`
+}
+
+// Please add new supporteed build-new-drivers structures here,
+// so that the utility starts building configs for them.
+// Fields must have the same name used in kernel-crawler json keys.
+type KernelList struct {
```
FedeDP, line 23: Another nice side-effect: we will only build configs for supported distros.
Note: this list should always match the `build-drivers` jobs configs, ie: the distro we want to support prebuilt artifacts for.

images/update-dbg/dbg-go/pkg/autogenerate/constants.go
```diff
@@ -0,0 +1,28 @@
...
+	urlLastDistro   = "https://raw.githubusercontent.com/falcosecurity/kernel-crawler/kernels/last_run_distro.txt"
+	configPathFmt   = "%s/driverkit/config/%s/%s/%s" // Eg: repo-root/driverkit/config/5.0.1+driver/x86_64/centos_5.14.0-325.el9.x86_64_1.yaml
+	outputPathFmt   = "output/%s/%s/%s_%s_%s_%s"     // Eg: output/5.0.1+driver/x86_64/falco_centos_5.14.0-325.el9.x86_64_1.{ko,o}
+)
+
+var (
+	// Please add new supported build-new-drivers structures here,
+	// so that the utility starts building configs for them.
+	// Fields must have the same name used in kernel-crawler json keys.
+	supportedDistros = []string{
```
FedeDP, line 14: This is the home of the supportedDistro by test-infra (ie: the distro for whom we provide prebuilt artifacts). Side effect: we only build configs for supported distros!

FedeDP:
Little benchmark, to generate correct configs for x86_64, driver version `5.0.1+driver`:
**dbg-go:**
```
time ./dbg-go autogenerate  --repo-root /home/federico/Work/test-infra/ --driver-version 5.0.1+driver INFO[0000] generating configs for AmazonLinux            cmd=autogenerate
INFO[0000] generating configs for AmazonLinux2           cmd=autogenerate
INFO[0000] generating configs for BottleRocket           cmd=autogenerate
INFO[0000] generating configs for CentOS                 cmd=autogenerate
INFO[0000] generating configs for Debian                 cmd=autogenerate
INFO[0000] generating configs for Minikube               cmd=autogenerate
INFO[0000] generating configs for Ubuntu                 cmd=autogenerate
INFO[0000] generating configs for AlmaLinux              cmd=autogenerate
INFO[0000] generating configs for AmazonLinux2023        cmd=autogenerate
INFO[0000] generating configs for Fedora                 cmd=autogenerate
INFO[0000] generating configs for Talos                  cmd=autogenerate
INFO[0000] generating configs for AmazonLinux2022        cmd=autogenerate

real    0m7,864s
user    0m12,067s
sys     0m1,571s

ls -l ../../../driverkit/config/5.0.1+driver/x86_64/ | wc -l
3770
```

**old bash scripts:**
```
time make TARGET_ARCH=x86_64 TARGET_VERSION=5.0.1+driver generate/auto
...
real    53m56,094s
user    46m44,610s
sys     6m59,271s

ls -l ../../../driverkit/config/5.0.1+driver/x86_64/ | wc -l
7039

ls -l ../../../driverkit/config/5.0.1+driver/x86_64/ubuntu-* | wc -l
2944

ls -l ../../../driverkit/config/5.0.1+driver/x86_64/amazonlinux* | wc -l
351

ls -l ../../../driverkit/config/5.0.1+driver/x86_64/centos_* | wc -l
283

ls -l ../../../driverkit/config/5.0.1+driver/x86_64/bottlerocket_* | wc -l
60

ls -l ../../../driverkit/config/5.0.1+driver/x86_64/talos_* | wc -l
16

ls -l ../../../driverkit/config/5.0.1+driver/x86_64/debian_* | wc -l
96

ls -l ../../../driverkit/config/5.0.1+driver/x86_64/minikube_* | wc -l
5

ls -l ../../../driverkit/config/5.0.1+driver/x86_64/fedora_* | wc -l
9

ls -l ../../../driverkit/config/5.0.1+driver/x86_64/almalinux_* | wc -l
5

SUM: 2944+351+283+60+16+96+5+9+5=3769
```

NOTE:
* in both i skipped the download json part to obtain better results.
* the actual SUM of the supported distros generated configs is the same, since `dbg-go` `ls -l` also contains the `total 101380` line.

Pretty good!

FedeDP:
This PR starts by removing `driverkit/utils/{scrape_and_generate,generate}` scripts, and their linked `{generate,generate/auto}` makefile's targets.
Everything is reimplemented in pure go.
The update-dbg image was updated to build and install the new tool, and the entrypoint script was updated to use it.

images/update-dbg/Dockerfile
```diff
@@ -1,7 +1,10 @@
-FROM golang:1.18 AS pullrequestcreator
+FROM golang:1.20 AS gobuilder
```
FedeDP, line 1: `pr-creator` now needs at least golang:1.19; given that golang:1.20 has glibc version that is too new to be used in debian:buster, compile with CGO disabled.
We will need to port this fix in all other images using `pr-creator` from test-infra.

images/update-dbg/dbg-go/pkg/autogenerate/autogenerate.go
```diff
@@ -0,0 +1,130 @@
...
+	jsonData, err := utils.GetURL(url)
+	if err != nil {
+		return err
+	}
+	logger.WithField("cmd", "autogenerate").Debug("fetched json")
+
+	// Generate a dynamic struct with all needed distros
+	// NOTE: we might need a single distro when `lastDistro` is != "*";
+	// else, we will add all supportedDistros found in constants.go
+	instanceBuilder := dynamicstruct.NewStruct()
```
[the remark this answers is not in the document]
maxgio92: Interesting @FedeDP!

images/update-dbg/dbg-go/pkg/autogenerate/constants.go
maxgio92: I think it makes sense @FedeDP!

images/update-dbg/dbg-go/pkg/autogenerate/autogenerate.go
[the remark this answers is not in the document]
FedeDP: Note: we might also unmarshal to a `map[string][]KernelEntry`, but then we would unmarshal any key (ie:  list of kernels for any distro) and then we would need to post filter `lastDistro` or supportedDistros.

Or, we might want to implement `Unmarshaler` interface perhaps?
Anyway, i think this is good to go, i find it quite simple to read and just works.

FedeDP:
Latest commit implements a `validate` subcmd that replaces ` utils/checkfiles` script.
Note: i need to move the `dbg-go` source files elsewhere, likely below driverkit/ folder, and then add some new triggers for the `update-dbg` image rebuild/republish when the driverkit/go-dbg folder gets modified.

FedeDP:
>  and then step by step first move driverkit Makefile's targets to use dbg-go behind the scenes, one after one.
When all the Makefile's targets use dbg-go, it will be time to drop the Makefile completely and update the README docs.

On a second thought, moving makefile's target implementation to use `dbg-go` tool is not as straightforward as it seems, because it then means that dbg Makefile depends upon the dbg-go script but we have no clear way to state that. I mean, we can document it, but given user needs to know that `dbg-go` exists and must be built before proceeding, i think it might be better to just drop the Makefile targets and update documentation pointing to the new tool.

FedeDP:
> Note: i need to move the dbg-go source files elsewhere, likely below driverkit/ folder, and then add some new triggers for the update-dbg image rebuild/republish when the driverkit/go-dbg folder gets modified.

Also, given that `validate` Makefile target is used by `build-drivers` image (presubmit jobs for drivers build), we will need to add the new dbg-go tool in the `build-drivers` image too to make use of it and drop the Makefile `validate` target.

IMHO, as a first step, i'd go with what we have today, ie: autogeneration and clenaup of old configs ported to `dbg-go` and `update-dbg` image only making use of the new tool.
Then we can calmly implement remaining targets in the go tool, 1 by 1. 
Finally, we will need to move the `dbg-go` tool in the driverkit folder, install it in the `build-drivers` image too, and add new build/publish triggers for both images on `dbg-go` tool sources updates, and enable its usage in the `build-drivers` entrypoint script. This will allow us to completely drop Makefile and remaining bash utils in dbg (and then rewrite the doc!)

images/update-dbg/dbg-go/pkg/autogenerate/autogenerate.go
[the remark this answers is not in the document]
maxgio92: We could implement the Unmarshaler, anyway I completely agree going with the Dynamic Structs 👍🏻

images/update-dbg/Dockerfile
```diff
@@ -1,20 +1,24 @@
...
 
 RUN git clone https://github.com/kubernetes/test-infra
-RUN cd test-infra/robots/pr-creator && go build -v -o pr-creator ./main.go
+RUN cd test-infra/robots/pr-creator && CGO_ENABLED=0 go build -v -o pr-creator ./main.go
 
 FROM debian:buster
 
+RUN wget -q https://github.com/FedeDP/dbg-go/releases/download/v0.4.0/dbg-go_0.4.0_linux_amd64.tar.gz \
```
FedeDP, line 8: Note: this will become `falcosecurity/dbg-go` once the repo gets donated.

images/build-drivers/build-drivers.sh
```diff
@@ -9,25 +9,22 @@ set -o errexit
...
-# DBG_WORKDIR: the absolute path from where to run DBG
 # ENSURE_DOCKER: whether to ensure Docker daemon running
 # DBG_MAKE_BUILD_TARGET: specify the makefile target
 
 # Needed Bash >= 4.0
 declare -A DBG_FILTERS
-DBG_FILTERS['TARGET_DISTRO']="${1}"
-DBG_FILTERS['TARGET_KERNEL']="${2}"
-DBG_FILTERS['TARGET_VERSION']="${3}"
-DBG_FILTERS['TARGET_ARCH']="${4:-$(uname -m)}"
```
FedeDP, line 21: Dropped unused filter.

images/build-drivers/build-drivers.sh
```diff
@@ -9,25 +9,22 @@ set -o errexit
...
+DBG_FILTERS['--target-distro']="${1}"
+DBG_FILTERS['--target-kernelrelease']="${2}"
+DBG_FILTERS['--target-kernelversion']="${3}"
+
 DBG_MAKE_BUILD_OPTIONS=""
-DBG_MAKE_BUILD_TARGET="${DBG_MAKE_BUILD_TARGET:-"specific_target"}"
+DBG_MAKE_BUILD_TARGET="${DBG_MAKE_BUILD_TARGET:-"build"}" # build or validate
```
FedeDP, line 22: Now the CLI subcmd names are either `build` or `validate`.

images/build-drivers/build-drivers.sh
```diff
@@ -59,29 +56,28 @@ function start_docker() {
...
-	touch output/failing.log
-	DBG_COMMIT=$(git log -1 --format=format:%H --full-diff -- ./)
-
-	pretty_echo "Found current driverkit version DBG_COMMIT $DBG_COMMIT. Running DBG build..."
-	$make $DBG_MAKE_BUILD_OPTIONS $DBG_MAKE_BUILD_TARGET
+	pretty_echo "Running DBG build..."
+	# when building, ignore errors (just report them back to output/failing.log)
+	if [[ "$DBG_MAKE_BUILD_TARGET" -eq "build" ]]; then
+		touch output/failing.log 
+		# TODO: support output redir in dbg-go
```
FedeDP, line 68: I still need to add this feat to `dbg-go`.

images/build-drivers/build-drivers.sh
```diff
@@ -59,29 +56,28 @@ function start_docker() {
...
+	fi
+	dbg-go configs $DBG_MAKE_BUILD_TARGET $DBG_MAKE_BUILD_OPTIONS
 	
 	pretty_echo "DBG build complete"
 }
 
 function publish() {
 	pretty_echo "Running DBG publishing..."
-	$make $DBG_MAKE_PUBLISH_TARGET
+	# TODO: pass aws-profile parameter
```
FedeDP, line 78: cc @maxgio92 do you have any idea how is the aws profile called?

images/build-drivers/build-drivers.sh
FedeDP: Will need to bump dbg-go to 0.0.5 (or 0.1.0).

images/build-drivers/build-drivers.sh
```diff
@@ -58,37 +54,27 @@ function start_docker() {
...
-	DBG_COMMIT=$(git log -1 --format=format:%H --full-diff -- ./)
-
-	pretty_echo "Found current driverkit version DBG_COMMIT $DBG_COMMIT. Running DBG build..."
-	$make $DBG_MAKE_BUILD_OPTIONS $DBG_MAKE_BUILD_TARGET
+	pretty_echo "Running DBG build..."
+	if [[ "$DBG_MAKE_BUILD_TARGET" -eq "build" ]]; then
+		# when building, ignore errors (just report them back to driverkit/output/failing.log)
+		DBG_MAKE_BUILD_OPTIONS="${DBG_MAKE_BUILD_OPTIONS} --ignore-errors --skip-existing --redirect-errors=driverkit/output/failing.log"
+		# If requested, publish too!
+		test "${PUBLISH_S3}" == "true" && DBG_MAKE_BUILD_OPTIONS="${DBG_MAKE_BUILD_OPTIONS} --publish"
```
FedeDP, line 68: `dbg-go` supports publishing while building (ie: asynchronously in a goroutine while the next build is being done).
This has 2 advantages:
* drivers will become available right after being built (without wait all the drivers to be built before start publishing)
* time spent uploading drivers is saved, since it is done async

FedeDP: > time spent uploading drivers is saved, since it is done async

Even if it does not seem that big of a deal, imagine if we take like 1s to upload a single driver; and imagine we upload like 2000 drivers each week (i haven't got any data about it atm); we are saving 2000 seconds, ie ~30mins!

FedeDP:
This is still wip because i still need to donate `dbg-go` repo!

config/jobs/build-drivers/build-new-almalinux.yaml
```diff
@@ -14,8 +14,6 @@ presubmits:
         - /workspace/build-drivers.sh
         - almalinux
         env:
-        - name: AWS_REGION
```
FedeDP, line 17: Do we want to retain this? Right now `dbg-go` specifies the region programmatically.

FedeDP:
Since `dbg-go` tool has been donated to falcosecurity, this PR is now ready.
/cc @maxgio92 
/hold

maxgio92:
Thank you @FedeDP for this incredible work!

images/build-drivers/build-drivers.sh
maxgio92: This is great

images/update-dbg/entrypoint.sh
```diff
@@ -30,10 +30,11 @@ export GIT_AUTHOR_NAME=${BOT_NAME}
 export GIT_AUTHOR_EMAIL=${BOT_MAIL}
 
 # Update the DBG from crawling results.
-# $2: chart name
 generate_driverkit_configs() {
-    make -C driverkit generate/auto
-    return $?
+    dbg-go configs cleanup -a amd64
```
maxgio92, line 34: @FedeDP why not doing for all architectures? Let's imagine we increase the architecture support

config/jobs/build-drivers/build-new-almalinux.yaml
maxgio92: @FedeDP does `dbg-go` set by default _eu-west-1_ region?

images/update-dbg/entrypoint.sh
FedeDP: Because `dbg-go` takes the architecture as input parameter, and does not support looping on architectures (that's because it needs to know that target arch while generating or building dbg configs).

maxgio92: No problem. Just my curiosity

maxgio92 (approved):
/approve

PR #1207, opened 2023-08-03
update(config.yaml): drop non-existent required job from falco

jasondellaluce:
The `test-dev-packages-arm64` is not existent when running workflows on PRs for Falco.


FedeDP (approved):
/approve

PR #1208, opened 2023-08-04
update(config.yaml): right naming from required Falco jobs (2)

jasondellaluce:
I'm struggling to figure out what the right nomenclature is for GitHub actions to properly match the required jobs by name. I tested these ones manually through the GH UI, so I hope this will be the last attempt.


leogr (approved):
/approve

PR #1224, opened 2023-09-01
Update (org.yaml): add rohith as a member 

Rohith-Raju:
See: https://github.com/falcosecurity/evolution/issues/303

### DIFF
--- a/config/org.yaml
+++ b/config/org.yaml
@@ -51,6 +51,7 @@ orgs:
       - terylt
       - vjjmiras
       - zuc
+      - rohith-raju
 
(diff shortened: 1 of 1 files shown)


config/org.yaml
```diff
@@ -51,6 +51,7 @@ orgs:
       - terylt
       - vjjmiras
       - zuc
+      - rohith-raju
```
FedeDP, line 54: Can you move it up, so that the list remains sorted? :)

FedeDP (approved):
/approve

PR #1227, opened 2023-09-04
update(org): add LucaGuerra to falcosecurity/falco maintainer group

LucaGuerra:
As per title. This way Poiana's commands should work properly. I am already in the OWNERS file for that repo.


FedeDP (approved):
/approve

Issue #1158, opened 2023-05-17
Drivers build not triggered for x86_64


maxgio92:
**Describe the bug**

Not all driver building-postsubmit [jobs](https://github.com/falcosecurity/test-infra/tree/master/config/jobs/build-drivers) have been triggered after new driverkit configurations [have been introduced](https://github.com/falcosecurity/test-infra/pull/1151) - as usual weekly crawling result from [kernel-crawler](https://github.com/falcosecurity/kernel-crawler).

**How to reproduce it**

Please check:
* [jobs for x86_64](https://prow.falco.org/?job=build-new*postsubmit) vs
* [jobx for arm64](https://prow.falco.org/?job=build-new*postsubmit-arm).

**Expected behaviour**

All the jobs for x86 to have been triggered accordingly.

FedeDP:
Note: all the jobs have always been instead triggered correctly during normal weekly kernel-crawler CI.
This time, instead, as we added new driver version (therefore copy/pasting an existing driver version to a new folder) and removed the oldest one (2.0.0+driver), not all jobs were triggered.

At the same time, PR CI correctly triggered all presubmit jobs instead.

FedeDP:
I might have found a good candidate; note the difference between the presubmit and postsubmit Prow documentation about `run_if_changed`:

**Pre:**
```
Presubmit jobs apply run_if_changed and skip_if_only_changed filters based on which files were modified in any of the commits in the pull request.
```

**Post:**
```
Postsubmit jobs apply run_if_changed and skip_if_only_changed filters based on which files were modified by the commits included in the specific push event from github.
```

So, postsubmit are based on the specific push event triggered by github. 
What if, given the large size of the diff/high number of files changed, the push event is somewhat truncated, ie: not all changed files are pushed?

Payloads are capped to 25M: https://docs.github.com/en/webhooks-and-events/webhooks/webhook-events-and-payloads#push 
> Note: Payloads are capped at 25 MB. If your event generates a larger payload, a webhook will not be fired. This may happen, for example, on a create event if many branches or tags are pushed at once. We suggest monitoring your payload size to ensure delivery.

But this is not our case imho because Prow receives the event with a full payload (otherwise it won't be able to decode it); but there might be some other hard limits on the number of files changes pushed with the event.

FedeDP:
So, github http api does limit list of files changed to 3000: https://docs.github.com/en/rest/pulls/pulls?apiVersion=2022-11-28#list-pull-requests-files

[...]

PR #1229, opened 2023-09-07
new(driverkit): add new driver version `6.0.0+driver` and remove the old `3.0.1+driver`

Andreagit97:
This PR adds new driver version `6.0.0+driver` and removes the old `3.0.1+driver`


FedeDP (approved):
NOTE: since we are still being hit by #1158  we will probably need multiple passes before all the drivers will be published.
Let's start asap so that we will achieve 100% coverage soon!
/approve

PR #1232, opened 2023-09-08
new: allow pushing to ECR falco-driver-loader-legacy and falco-distroless

LucaGuerra:
For Falco 0.36.0 we plan to distribute two new images:
* falco-driver-loader-legacy which will contain the old driver loader, for compatibility with older kernel
* falco-distroless which will contain an experimental minimal image to reduce attack surface

Add those to ECR


FedeDP (approved):
/approve

PR #1233, opened 2023-09-08
new(tf): add Falco ECR public repositories

LucaGuerra:
The Terraform distribution configuration was missing entries for `falco` and `falco-driver-loader`.

In addition, add the following images:
* falco-driver-loader-legacy for the older driver loader image
* falco-distroless for the experimental distroless/minimal image

### DIFF
--- a/config/distribution/containers.tf
+++ b/config/distribution/containers.tf
@@ -39,3 +39,93 @@ resource "aws_ecrpublic_repository" "falcosidekick_ui" {
     prevent_destroy = true
   }
 }
+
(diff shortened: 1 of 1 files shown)


FedeDP (commented):
LGTM!
Only concern: is `ARM 64` the correct value?

FedeDP:
It is! https://docs.aws.amazon.com/it_it/AmazonECS/latest/developerguide/firelens-using-fluentbit.html

FedeDP (approved):
/approve

PR #1236, opened 2023-09-12
chore(config/jobs): workaround #1158

FedeDP:
Proposed way to workaround #1158 , even if this is not a real fix: trigger all build-drivers jobs on any new config.
See https://github.com/falcosecurity/test-infra/issues/1158#issuecomment-1710018485

PROs:
* super easy workaround

CONs:
* we will spawn useless pods
* we will spawn more nodes than needed

However, please note that:
* nowadays, since kernel-crawler automation was implemented, we rarely have PRs with a single config being added (it did not happen in 1yr)
* normally, we trigger the build of drivers when: either the kernel-crawler automation triggers, or we add a new driver version; those jobs normally trigger most of our jobs (~90/100%), therefore spawning all pods in this case would make no real difference
* Only case where we would scale badly is when we manually run the kernel-crawler automation against a single distro; in this case, instead of only triggering the distro build-drivers jobs, we would trigger all (even if `skip existing` prevent us from rebuilding everything of course)

In this PR, i trigger indifferently x86_64 and aarch64 jobs, since when eg: we add a new driver version, we basically only trigger aarch64 jobs before the Github API cuts our list of changed files. 
In that case, to actually trigger x86_64 jobs, we need this fix.

### DIFF
--- a/config/jobs/build-drivers/build-new-almalinux.yaml
+++ b/config/jobs/build-drivers/build-new-almalinux.yaml
@@ -48,7 +48,7 @@ postsubmits:
     agent: kubernetes
     branches:
       - ^master$
-    run_if_changed: '^driverkit/config/[a-z0-9.+-]{5,}/x86_64/almalinux_.+'
+    run_if_changed: '^driverkit/config/[a-z0-9.+-]{5,}/' #x86_64/almalinux_.+'
     spec:
       serviceAccountName: driver-kit
       containers:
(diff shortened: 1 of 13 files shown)


maxgio92:
@FedeDP see https://github.com/falcosecurity/test-infra/issues/1158#issuecomment-1719314556

FedeDP:
Let's close this one: with daily kernel-crawler jobs, this issue is less and less important.
/close

PR #1249, opened 2023-09-26
chore(config): protect dbg-go main branch.

FedeDP:


maxgio92 (approved):
Good catch @FedeDP!

/approve

PR #1250, opened 2023-09-27
update(distribution): vendor s3-bucket-listing

LucaGuerra:
The current version of the s3-bucket-listing script does not display our bucket properly. Therefore we are going to vendor the previous commit in our repository, retaining the copyright notice of course.


LucaGuerra:
This is ready! Fixes the download problem.

FedeDP (approved):
/approve

PR #1260, opened 2023-10-19
chore(config/jobs): move update-dbg to cron type.

FedeDP:
/hold

Following https://github.com/falcosecurity/kernel-crawler/pull/179 discussion, moving `update-dbg` to periodic trigger.
NOTE: kernel-crawler job will run at 6.30 UTC, and the job lasts around 35minutes. This job is triggered at 9 UTC, we have plenty of time :)


maxgio92 (approved):
/approve

PR #1262, opened 2023-10-19
fix status check

cpanato:
follow up #1261

looks like needs to be the full name


leogr (approved):
/appreve

cpanato:
hope this fixes

PR #1265, opened 2023-10-19
update(config): unprotect charts gh-pages branch

LucaGuerra:
The branch was protected before and allowed admins (the bot) to push to it. Now we are migrating that job to GitHub Actions, which cannot push to protected branches (https://github.com/orgs/community/discussions/25305). This allows the action to work.

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -85,12 +85,6 @@ branch-protection:
           branches:
             master:
               protect: true
-            gh-pages:
-              protect: true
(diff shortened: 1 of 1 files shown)


leogr (approved):
IMO, this is a temporary workaround to solve the current problem we have in the charts repository. However, not having the `gh-pages` branch protected is dangerous. 

So I'm approving this, but  I would like to find a better solution as soon as possible.

/approve

cc @falcosecurity/test-infra-maintainers @falcosecurity/charts-maintainers

cpanato:
lets have a meeting to discuss this and options

PR #1267, opened 2023-10-20
fix(images/build-drivers): fixed drivers validation step.

FedeDP:


maxgio92 (approved):
/approve

Good catch @FedeDP!

PR #1271, opened 2023-10-23
cleanup: single validate job

FedeDP:
It is nowadays useless to spin up 30pods to just validate dbg configs; just create a single job that validates all dbg configs.
Moreover, fixed a bug in `build-drivers.sh` script that was only validating x86_64 dbg configs.


FedeDP:
/hold 
What to do with `build-drivers.local` file? It just has a fake presubmit job.

config/jobs/build-drivers/validate-dbg.yaml
FedeDP, line 1: I put it in the same folder because it uses same `build-drivers` image and has a similar trigger.

images/build-drivers/build-drivers.sh
```diff
@@ -55,26 +55,36 @@ function start_docker() {
...
+	DBG_MAKE_BUILD_OPTIONS="${DBG_MAKE_BUILD_OPTIONS} --ignore-errors --skip-existing --redirect-errors=driverkit/output/failing.log"
+	# If requested, publish too!
+	test "${PUBLISH_S3}" == "true" && DBG_MAKE_BUILD_OPTIONS="${DBG_MAKE_BUILD_OPTIONS} --publish"
+	dbg-go configs build $DBG_MAKE_BUILD_OPTIONS
+
+	pretty_echo "dbg-go build complete"
+}
+
+function validate() {
+	dbg-go configs validate --architecture arm64
```
FedeDP, line 75: Validate arm64 configs too! This is a fix.

FedeDP:
/unhold
let's keep it there.

maxgio92:
> /hold What to do with `build-drivers.local` file? It just has a fake presubmit job.

They're the job configs for local Prow, useful to test jobs locally (more [here](https://github.com/falcosecurity/test-infra/blob/08b3865c9d8cd602a1636bc792aa8766fa56e8fc/docs/local-testing.md#local-prow-testing))

images/build-drivers/build-drivers.sh
maxgio92: Great catch @FedeDP

maxgio92:
I like it, and it makes sense now with `dbg-go` as it's one order of magnitude more efficient

images/build-drivers/build-drivers.sh
maxgio92: Can we make explicit the validation also for x86? So that, when running on arm64 nodes will run not only on arm

FedeDP: Done!

maxgio92 (approved):
/approve

PR #1316, opened 2023-12-11
cleanup(config,tools): drop ARC.

FedeDP:
We now use actuated nodes for github action arm64 runners: https://github.com/falcosecurity/falco/pull/2945


maxgio92 (approved):
/approve

maxgio92:
As we use arm64 nodes only for GH runner, we can now remove them too.

FedeDP:
Nope we use them also for the drivers build!

PR #1329, opened 2023-12-18
chore(config): updated driverkit required status checks.

FedeDP:


FedeDP:
This should allow merging https://github.com/falcosecurity/driverkit/pull/311

PR #1381, opened 2024-02-15
update(config): temporarily remove required from asan build

LucaGuerra:
This is in preparation for https://github.com/falcosecurity/libs/pull/1685 . That PR changes the name of the ASan job (which becomes `sanitizers`) and adds its ARM64 counterpart. After libs#1685 is merged we'll be able to add the new jobs as required.


maxgio92 (approved):
/approve

PR #1389, opened 2024-02-20
fix(tf): reduce falcoctl about field length

LucaGuerra:
I believe this wasn't applying cleanly because it was counting the bytes and we had emojis in the text. I removed a few more bytes to make sure.


FedeDP (approved):
/approve

PR #1415, opened 2024-03-12
fix(tf): add S3 list access to Falco Playground

LucaGuerra:
This allows the release job on Falco Playground to list the bucket contents during the sync operation


FedeDP (approved):
/approve

PR #1428, opened 2024-03-25
update(org.yaml): adding sboschman to falcosecurity/plugins/k8saudit-gke

leogr:
See https://github.com/falcosecurity/plugins/pull/437

cc @FedeDP


FedeDP (approved):
/approve

PR #1432, opened 2024-03-28
update(config/config.yaml): switch default branch to `main` for `plugins` repo

leogr:
/hold until we switch the default branch in falcosecurity/plugins

### DIFF
--- a/config/config.yaml
+++ b/config/config.yaml
@@ -336,7 +336,7 @@ branch-protection:
               - "build-and-test"
         plugins:
           branches:
-            master:
+            main:
               protect: true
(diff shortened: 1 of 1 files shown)


FedeDP (commented):
Most probably we will also need to bump iam roles config: https://github.com/falcosecurity/test-infra/blob/master/config/clusters/iam.tf#L197

FedeDP:
I think https://github.com/falcosecurity/test-infra/blob/master/config/jobs/build-plugins/build-plugins.yaml#L8 must be updated too.

FedeDP (approved):
/approve

PR #1440, opened 2024-04-08
fix(config/jobs/update-github-teams): set correct trigger branch

maxgio92:
This PR fixes errors on the update-github-teams jobs trigger branches, that erroneously were set to _main_.


FedeDP (approved):
Thanks max!
/approve

PR #1444, opened 2024-04-11
chore(config/jobs): add test peribolos-syncer presubmit

maxgio92:
This is a temporary test for presubmit to test trigger


FedeDP (approved):
Let's try ...
/approve

PR #1455, opened 2024-04-12
update(config/jobs): bump peribolos-syncer to v0.2.2.

FedeDP:
/hold
We still need to understand why jobs are not triggering.

@maxgio92


maxgio92 (approved):
/approve

PR #1456, opened 2024-04-12
cleanup(config): remove chisel jobs no more required

therealbobo:
This removes the "Required" mark from the chisel jobs given that the chisel will be deprecated. https://github.com/falcosecurity/libs/issues/1743


FedeDP (approved):
/approve

PR #1487, opened 2024-05-08
chore(config): cleaning up unexisting user

leogr:
The following account (of an emeritus maintainer) no longer exists, and https://github.com/markyjackson-taulia doesn't either. So, thus PR is just cleaning it up.

Note: Marky, if you read this, I wanted just to let you know that you will always be welcome here :hugs:


FedeDP (approved):
/approve

PR #1528, opened 2024-06-20
chore(ci): switch to kubeconform since kubeval is deprecated.

FedeDP:
Kubeval is no longer maintained since 2021: https://github.com/instrumenta/kubeval
Use the proposed replacement tool: https://github.com/yannh/kubeconform


maxgio92 (approved):
Thank you @FedeDP! Good catch!

These pipelines need definitely some care  🤗

PR #1558, opened 2024-07-16
new(config/prow): add pod-identity-webhook

alacuku:
Now the service accounts can be annotated with iam roles in order to get the right permissions.


FedeDP (approved):
/approve

PR #1637, opened 2024-09-25
update(config): (re)move zig

LucaGuerra:
Due to a necessary patch release in libs we need to disable required checks for zig builds since that is not enabled on the release branch. Partially reverts https://github.com/falcosecurity/test-infra/pull/1635 (we know what we doing).


FedeDP (approved):
/approve

FedeDP:
Thanks luca!

PR #1682, opened 2024-11-07
update(cert-manager): introduce argocd application for cert-manager

alacuku:
Remove the cert-manager manifest and update the application version to 1.16.1


FedeDP (approved):
/approve

PR #1684, opened 2024-11-08
update(alb-controller): introduce argocd application for alb-controller

alacuku:
Furthermore, bump version to 2.10.0


FedeDP (approved):
/approve

PR #1703, opened 2024-11-21
new(app): prometheus operator crds

alacuku:
This commit introduces a new argocd application that handles CRDS for prometheus operator.


FedeDP (approved):
/approve

PR #1737, opened 2025-01-09
Add driver for new hwe ubuntu generic kernel

gbittenbender:
Was receiving 404 errors from attempting to use a ubuntu-generic HWE kernel. Similar kernel drivers appear to have been built for previous versions, so opening this pull request to create one for my desired kernel

### DIFF
--- a/driverkit/config/7.3.0+driver/x86_64/ubuntu-generic_5.15.0-124-generic_134~20.01.1.yaml
+++ b/driverkit/config/7.3.0+driver/x86_64/ubuntu-generic_5.15.0-124-generic_134~20.01.1.yaml
@@ -1,0 +1,7 @@
+kernelversion: 134~20.01.1
+kernelrelease: 5.15.0-124-generic
+target: ubuntu-generic
+architecture: amd64
+output:
+    module: output/7.3.0+driver/x86_64/falco_ubuntu-generic_5.15.0-124-generic_134~20.01.1.ko
+    probe: output/7.3.0+driver/x86_64/falco_ubuntu-generic_5.15.0-124-generic_134~20.01.1.o
--- a/driverkit/config/7.3.0+driver/x86_64/ubuntu-generic_5.15.0-130-generic_140~20.01.1.yaml
+++ b/driverkit/config/7.3.0+driver/x86_64/ubuntu-generic_5.15.0-130-generic_140~20.01.1.yaml
@@ -1,0 +1,7 @@
+kernelversion: 140~20.01.1
+kernelrelease: 5.15.0-130-generic
+target: ubuntu-generic
+architecture: amd64
+output:
+    module: output/7.3.0+driver/x86_64/falco_ubuntu-generic_5.15.0-130-generic_140~20.01.1.ko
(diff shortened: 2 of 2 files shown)


FedeDP:
Hi! Thanks for opening this PR! Your commit is missing DCO.

Also, that's weird: we only crawl https://falcosecurity.github.io/kernel-crawler/?arch=x86_64&target=ubuntu&search=5.15.0-124-generic, but cannot find the `134~20.01.1` one. Can you share the ubuntu release?

gbittenbender:
> Hi! Thanks for opening this PR! Your commit is missing DCO.
> 
> Also, that's weird: we only crawl https://falcosecurity.github.io/kernel-crawler/?arch=x86_64&target=ubuntu&search=5.15.0-124-generic, but cannot find the `134~20.01.1` one. Can you share the ubuntu release?

Added the DCO. 

And yeah, it should be a pretty standard Ubuntu desktop.
image

FedeDP:
Mmmh from your screenshot, i'd expect the added kernel to be:
> 5.15.0-130-generic
> 140~20.04.1

gbittenbender:
> Mmmh from your screenshot, i'd expect the added kernel to be:
> 
> > 5.15.0-130-generic
> > 140~20.04.1

Oh good catch! We found this while working on updating a new image, so I must have added the info just from the past version. I'll update since I expect we'll need both.

FedeDP:
Thanks! Also, can you share your `/etc/apt/sources.list` file? I would love to understand where is this kernel coming from to fix the ubuntu crawler at https://github.com/falcosecurity/kernel-crawler/blob/main/kernel_crawler/ubuntu.py#L23

driverkit/config/7.3.0+driver/x86_64/ubuntu-generic_5.15.0-124-generic_134~20.01.1.yaml
```diff
@@ -0,0 +1,7 @@
+kernelversion: 134~20.01.1
```
FedeDP, line 1: Isn't this `134~20.04.1` ?
I found this one: https://ubuntu.pkgs.org/20.04/ubuntu-updates-main-amd64/linux-image-5.15.0-130-generic_5.15.0-130.140~20.04.1_amd64.deb.html

gbittenbender:
> If that's ok for you, i'd let the other PR get merged and tomorrow morning our automated workflow will crawl the new kernels and then build drivers against them and publish them.

Yup, that works for me. Thanks!

FedeDP:
Hey, I Just merged #1739 with the latest crawled kernels.
The drivers should be up in a couple of hours (time to build and publish them!)

Thanks again for reporting the missing driver!
/close

FedeDP:
Oh i forgot that we don't support `ubuntu-hwe` prebuilt drivers :/ 
We only support:
* ubuntu-aws
* ubuntu-gke
* ubuntu-azure
* ubuntu-generic
* ubuntu-gcp

After all, if you are on 5.15, you can use the modern ebpf driver that does not require any external driver download since it is embedded in the Falco binary itself (indeed, latest Falco should already default at it).

PR #1826, opened 2025-04-18
update(falco): use 0.41.0 preview of falco

alacuku:
Testing the latest features of the container and k8smeta plugins.


FedeDP (approved):
/approve

PR #1905, opened 2025-09-17
update(members): add Tero Kauppinen (terror96) as member of libs.

terror96:
Required by https://github.com/falcosecurity/libs/pull/2631.


maxgio92 (approved):
/approve

PR #1997, opened 2026-04-20
chore(images)!: bump `dbg-go` to `0.18.0`

ekoops:
Notice that `dgb-go@0.18.0` dropped support for generating config for building legacy eBPF probe drivers.


leogr (approved):
/approve

PR #1998, opened 2026-04-20
update(config): archiving client-go

leogr:
See https://github.com/falcosecurity/evolution/issues/494

cc @ekoops


ekoops (approved):
/approve

PR #2015, opened 2026-05-13
update(config/jobs): generalize chart sync job

c2ndev:
## What this PR does

Generalizes the Falco Operator chart sync automation into a reusable `sync-charts` updater.

The Falco Operator postsubmit remains the first consumer, but the image and entrypoint are now chart-agnostic and can be reused later for other source-owned charts.

## Notes

The updater keeps the source repository as the chart source of truth. It only syncs chart files into `falcosecurity/charts`; it does not render, lint, regenerate docs, or mutate chart content.

It also adds guardrails for:
- target path restricted to `charts/<chart>`
- source chart name matching the target chart directory
- required chart layout used by current Falco charts
- chart changes requiring a `Chart.yaml` version bump
- idempotent no-op when there are no changes
- PR title format: `sync(charts/<chart>): v<version>`

### DIFF
--- a/config/jobs/build-prow-images/build-images.yaml
+++ b/config/jobs/build-prow-images/build-images.yaml
@@ -196,12 +196,12 @@ presubmits:
           privileged: true
       nodeSelector:
         Archtype: "x86"
-  - name: build-images-update-falco-operator-chart
+  - name: build-images-sync-charts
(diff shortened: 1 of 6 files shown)


leogr (approved):
/approve

PR #2016, opened 2026-05-14
fix(config): scope prempti required status checks to main branch only

leogr:
See https://github.com/falcosecurity/prempti/pull/32


ekoops (approved):
/approve

PR #2020, opened 2026-05-15
fix(sync-charts): reuse PRs from branch names with slashes

c2ndev:
## What this PR does

Fixes the `sync-charts` updater so `pr-creator` can reuse an already open chart sync PR when the generated branch name contains slashes.

The job already pushes to a branch in the target repository, for example:

```text
sync/charts/falco-operator
```

However, `pr-creator` was searching with an owner-qualified head branch:

```text
head:falcosecurity:sync/charts/falco-operator
```

GitHub search does not match that form for branch names with slashes, so the updater could push the commit but then fail while trying to create a duplicate PR.


leogr (approved):
/approve

ekoops (approved):
/approve

PR #2024, opened 2026-05-20
chore(prow): remove peribolos-syncer jobs and add falco chart sync

c2ndev:
## What this PR does

Removes the deprecated `peribolos-syncer-*` Prow jobs under `config/jobs/update-github-teams`.

Those jobs only attempted to generate `config/org.yaml` sync PRs from repository `OWNERS` files. They are currently stale/broken and are not the Peribolos jobs that apply `config/org.yaml` to GitHub.

This also adds the missing `sync-charts-falco` postsubmit, aligned with the existing chart sync jobs for `falco-operator`, `event-generator`, and `k8s-metacollector`.


ekoops (approved):
/approve

PR #2029, opened 2026-05-22
prow: sync Falco chart from release branches

c2ndev:
**What this PR does / why we need it**:

Updates the Falco chart sync job to follow the chart release branch workflow.

The Falco chart sync now runs from `release/*` branches instead of `master`, so unreleased chart development can continue on `master` without publishing a chart.

The sync script can now derive the source chart URL from Prow-provided repository metadata when `SOURCE_REPO_URL` is not set. Existing jobs that still set `SOURCE_REPO_URL` keep the current behavior.


ekoops (approved):
/approve

PR #2048, opened 2026-07-16
feat(config/clusters/oci): add OCI/OKE Prow platform and bootstrap stacks

c2ndev:
## Summary

Add an OCI/OKE platform alongside AWS/EKS, with independent Prow configuration,
node scheduling and deployment automation. Prow job migration and CI cutover are
outside this PR.

## Changes

- Move AWS cluster, application, Prow and job configuration into cloud-specific
  directories and update their consumers.
- Add OCI networking, OKE, image repositories and log storage, with node pools
  for the Prow platform, automation and x86/ARM DriverKit workloads.
- Keep IAM and state-bucket administration in a separate, maintainer-run
  Terraform bootstrap. Bind autoscaler permissions to its OKE workload identity.
- Add GitHub-hosted OCI Terraform automation using OIDC-federated temporary
  credentials, reviewed PR plans, identity verification and gated automatic apply.
- Add OCI Argo CD applications, monitoring, Gateway API routing and TLS, plus
  an independent Prow installation and Falco-branded dashboard.
- Select AWS, OCI or both validation workflows from changed paths, with separate
  Prow checks, manifest validation and Terraform security scans.
- Serialize AWS deployments, apply validated saved Terraform plans and update
  the affected image-build dependency pins. Retire the automatic AWS job uploader
  while retaining the manual tool.

## Deployment requirements

- Update and verify the running AWS config-updater mappings before merging the
  relocated paths. Preserve its ownership of existing configuration ConfigMaps.
  See the [AWS deployment instructions](https://github.com/falcosecurity/test-infra/blob/e42ee3bb622a8086beb312f28cfcb8e9457c97cb/tools/ci/README.md#automatic-aws-terraform-apply).
- Provision OCI application credentials externally. For an existing autoscaler,
  create its workload policy before switching authentication; verify scaling
  before removing the former worker grants. See the
  [application prerequisites](https://github.com/falcosecurity/test-infra/blob/e42ee3bb622a8086beb312f28cfcb8e9457c97cb/config/applications/oci/README.md) and
  [IAM transition](https://github.com/falcosecurity/test-infra/blob/e42ee3bb622a8086beb312f28cfcb8e9457c97cb/config/clusters/oci/bootstrap/README.md#transition-from-instance-principals).
- Keep automatic OCI Terraform apply disabled until federation and authorization
  checks pass. The OCI Prow bundle does not activate a job catalog or merge
[...]

PR #2066, opened 2026-09-11
ci(aws): validate Prow config across repository layouts

c2ndev:
Prepare the existing AWS check-prow-config presubmit and periodic for the directory moves in #2048. The active checker currently reads the legacy paths and fails when checking the new layout.

Select the AWS paths from the checked-out layout, reject mixed layouts, and propagate validation failures without falling back to another configuration. Keep the same checkconfig image, job names, scheduling, reporting, and node selector. Only the checker commands change.

Validation: 14 local cases passed with the existing AWS checkconfig image, pinned by digest for testing, without network access. Coverage includes both layouts, both job commands, invalid core/plugins/jobs, incomplete or mixed moves, and exclusion of the OCI job directory.

This is preparatory compatibility: #2048 must retain direct AWS paths and remove this selection logic when the layout changes. The config_updater path handoff remains a separate pre-merge step.

Kept as draft pending the production merge decision: merging invokes the existing AWS Prow deployment workflow and job-config update postsubmit. This PR does not change Terraform, workflows, secrets, or config_updater mappings.


alacuku (approved):
/lgtm

PR #2067, opened 2026-09-14
fix(prow): require the aggregate Chainsaw e2e check for falco-operator

c2ndev:
/kind bug

The falco-operator branch-protection configuration still requires `tests-e2e`, whose workflow is removed by https://github.com/falcosecurity/falco-operator/pull/418. Replace it with the stable aggregate check `tests-e2e-chainsaw`, added in that PR.

The aggregate runs after the binary build, the full Chainsaw matrix, and CRD validation. It requires all three job results to succeed, including when dependencies fail, are cancelled, or are skipped. Adding tests or matrix categories will therefore no longer require changes to test-infra. Existing lint, unit-test, and inherited DCO requirements are unchanged.

Wait for the new aggregate check to pass on falcosecurity/falco-operator#418, then merge and apply this configuration before merging that PR, so branch protection no longer waits for the removed check.

Validation:

- Passed the CI-pinned Prow `checkconfig` (`v20240805-37a08f946`) against the core configuration, jobs, and plugins.
- Matched the required context to the aggregate job name in falcosecurity/falco-operator#418.
- Validated the companion workflow with `actionlint` and exercised its result check across all 64 combinations of success, failure, cancellation, and skipping; only all-success passes.
- Parsed and compared the full YAML against `master`, confirming that only the falco-operator required e2e context changed; `git diff --check` passed.


leogr (approved):
/approve

PR #2070, opened 2026-09-15
fix(terraform): align AWS Kubernetes versions and OCI autoscaler inputs

c2ndev:
## Changes

- Separate the AWS EKS control-plane and managed-node-group Kubernetes versions. Set the control-plane default to the deployed version, 1.31, while keeping all three node groups at 1.30. Update the input documentation.
- Set the existing OKE cluster OCID in the manual bootstrap inputs so a regular bootstrap plan retains the autoscaler workload identity policy without requiring a command-line override. GitHub OIDC remains disabled.

## Validation

- AWS and OCI bootstrap Terraform formatting and validation pass.
- A local read-only AWS plan using the corrected sources reports no remote resource changes and no output changes. The control plane and all three node groups are unchanged; only local kubeconfig file creation remains.
- No apply was performed. OCI bootstrap IAM remains outside CI; this change does not activate OCI jobs or automatic Terraform apply.


alacuku (approved):
/lgtm

PR #2073, opened 2026-09-16
fix(oci): finalize Prow reconciliation and Terraform access

c2ndev:
- Enable server-side diff for the OCI Prow application.
- Fix Terraform federation and image discovery permissions.
- Remove internal operational documentation.

Validated OCI manifests, Prow configuration, Terraform and workflows locally. [Terraform verification](https://github.com/falcosecurity/test-infra/actions/runs/35111573049) completed with no changes for both identities.

No Prow jobs are migrated.


leogr (approved):
/approve

🥳

PR #2077, opened 2026-09-17
feat(prow): migrate the driver build grid to OCI

c2ndev:
Move all 71 driver publishing jobs, config validation and the scheduled grid updater to OCI. Preserve the original AWS job definitions as inactive backups.

- Keep job names, architecture, distribution and kernel filters.
- Expand DriverKit node filesystems and size autoscaler storage estimates accordingly.
- Use pinned upstream build tools and publish the updater image through GitHub Actions.
- Enable OCI job triggering and reporting while retaining merge management on AWS.

All CI checks pass, including both updater image builds. The OCI Terraform plan reports 0 additions, 3 in-place updates and 0 destructions. Prow configuration, rendered manifests, server-side dry-run and local updater execution also passed. Updater credentials are provisioned.

The merged workflow publishes the GHCR image; make it public before the first scheduled updater run at 08:00 UTC. Enable the OCI webhook only after Terraform apply, Argo synchronization and removal of the migrated jobs from the active AWS catalog have been verified. Native OCI build and publishing validation remains part of cutover.


leogr (approved):
/approve

PR #2078, opened 2026-09-17
fix(prow): preserve OCI trigger trust and centralize ownership

c2ndev:
Restrict the OCI trigger to organization members and preserve the contribution link. Plugin enablement remains limited to falcosecurity/test-infra.

Remove all four nested OWNERS files, leaving root OWNERS unchanged. Their approvers and emeritus approvers are already listed at the root; the nested reviewer has been inactive for over six months.


leogr (approved):
/approve

